Issue #2, opened 2024-08-24
[CWM] Create Swagger spec for MykoMap API


ms0ur1s:
### Description
Based on the API design for MykoMap discussed in [Issue DigitalCommons/mykomap#267](https://github.com/DigitalCommons/mykomap/issues/267) and the  [notes from the meeting on 20/8](https://docs.google.com/document/d/1aguAEMKfr8gqZRVkhJnb46VRJKd5wPMLLOQ1UmAKgFo/), create a Swagger (OpenAPI) specification. This spec will outline the API endpoints, request/response formats, and any necessary validation rules.

### Objectives

- Define the API endpoints as outlined in the meeting notes.
- Specify the HTTP methods (GET, POST, DELETE, etc.) for each endpoint.
- Include request and response schemas, including data models.
- Document error handling and expected status codes.

### Acceptance Criteria

- A complete Swagger spec that covers all discussed API endpoints.
- The spec should be validated and ready for use by the development team.
- Ensure the spec is reviewed and approved by at least one senior developer.

wu-lee:
I suggest we want to add some information about the API on the default route, or perhaps under `about/`

- [http://blah/api/about]() -> `{name: 'Mykomap API', apiVersion: 'v4.0.0:0fdeaf', deployed:'2024-09-02T14:56:04'}` (version is generated using something like `git describe --always --tags --dirty --abbrev=6`)

wu-lee:
> @wu-lee has created a draft API spec here: https://app.swaggerhub.com/apis/SWAGGERHUBWULEE/mykomap/0.1.0


Note this has moved to a DCC account here:

https://app.swaggerhub.com/apis/DigitalCommonsCoop/Mykomap/0.1.0

(Details in BitWarden)

Issue #1, opened 2024-08-24
[CWM] Create and deploy MM NodeJS server


rogup:
### Description
Based on the whiteboard https://github.com/DigitalCommons/mykomap/issues/266 we need to create a simple NodeJS backend server for MM.

### Objectives

- Create a 'hello world' NodeJS app in the `apps/back-end/` directory of the monorepo
- Deploy this app to dev-2
- Once this is working, use https://www.npmjs.com/package/fastify-openapi-glue to generate Fastify routes from the openapi spec that is created by https://github.com/DigitalCommons/mykomap-monolith/issues/2
- Serve mock responses for these routes

### Acceptance Criteria

- A working NodeJS/Fastify back-end that serves mock responses for each API route

PR #3, opened 2024-08-30
Add Node back-end and OpenAPI spec

rogup:
#### What? Why?

Related to issues #1 and #2

#### What should we test?

[edit: wu-lee]
-  Run `npm run test` in the apps/back-end directory. These are fairly rudimentary for now but should pass.

#### Deployment notes

[edit: wu-lee]
Deployment is still Work In Progress, and a subject of a branch which was split out of this one.

#### Documentation updates

[edit: wu-lee]
I don't think so...


apps/back-end/.editorconfig
```diff
@@ -0,0 +1,20 @@
+# Cross-platform formatting config
```
rogup, line 1: We can remove this file since it exists in the root directory

apps/back-end/package-lock.json
```diff
@@ -0,0 +1,2327 @@
+{
```
rogup, line 1: There shouldn't be a package-lock.json here, just one in the root directory which npm workspaces manages

apps/back-end/.editorconfig
wu-lee: This was added when it was a stand-alone project. So I'm ok removing this, for now at least.

But more generally, I was wondering what happens any of these subprojects need to become git-subtree inclusions later, so have their own self-standing repos.  It seems like those repos may have lost important definitions in the interest of de-duplication. Something to be wary of, a point Victor DevOps-Toolkit makes here about [monorepos encouraging [good/bad] tight coupling](https://youtu.be/VvcJGjjEyKo?t=499).

apps/back-end/package-lock.json
wu-lee: Ditto, see above.

rogup: This file can be generated by the package.json if we need to separate the project out, but I think it will be confusing for development and interfere with npm workspaces if we keep it here whilst in a monorepo

apps/back-end/.editorconfig
rogup: Yeah I think it makes sense not tightly couple the repos unnecessarily, especially in the actual code. I'm just thinking for files like this, which are mainly there for our developability, there's a risk of having duplicated files in different places that could get out of sync and lead to different codestyles.

PR #5, opened 2024-09-02
Add deploy script

rogup:
#### What? Why?

Related to issue # 

#### What should we test?

- Go to '...'
- 

#### Checklist

- [ ] Added UTs

### DIFF
--- a/apps/back-end/package.json
+++ b/apps/back-end/package.json
@@ -9,6 +9,7 @@
   "scripts": {
     "build": "vite build --ssr src/server.js --outDir dist && cp pm2/* dist",
     "start": "fastify start --options src/index.js",
+    "start:attached": "node src/server.js",
     "test": "c8 node --test test/test-*.js",
     "dev": "fastify start -l debug -P --options src/index.js"
   },
--- a/apps/back-end/package.json
+++ b/apps/back-end/package.json
@@ -7,7 +7,7 @@
   "private": true,
   "type": "module",
   "scripts": {
-    "build": "vite build --ssr src/server.js --outDir dist && cp pm2/ecosystem.config.js dist",
+    "build": "vite build --ssr src/server.js --outDir dist && cp pm2/* dist",
     "start": "fastify start --options src/index.js",
     "test": "c8 node --test test/test-*.js",
     "dev": "fastify start -l debug -P --options src/index.js"
--- a/apps/back-end/package.json
+++ b/apps/back-end/package.json
@@ -7,7 +7,7 @@
   "private": true,
   "type": "module",
   "scripts": {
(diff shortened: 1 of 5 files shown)


apps/back-end/package.json
```diff
@@ -7,8 +7,9 @@
   "private": true,
   "type": "module",
   "scripts": {
-    "build": "vite build --ssr src/server.js --outDir dist",
+    "build": "vite build --ssr src/server.js --outDir dist && cp pm2/* dist",
     "start": "fastify start --options src/index.js",
+    "start:attached": "node src/server.js",
     "test": "c8 node --test test/test-*.js",
     "dev": "fastify start -l debug -P --options src/index.js"
```
rogup, line 14: Note I changed this to get it working with the dev FE. I used LX's backend package.json for inspiration, which generally works well for dev/prod and we might want to copy.

deploy.sh
```diff
@@ -0,0 +1,130 @@
...
+mkdir -p "${SYSTEMD_UNIT%/*}"
+cat >"$SYSTEMD_UNIT" <<EOF
+[Unit]
+Description=Mykomap back-end process manager for %u
+Documentation=https://github.com/DigitalCommons/mykomap-monolith/
+After=network.target
+
+[Service]
+Type=exec
+LimitNOFILE=infinity
```
rogup, line 61: The unit file for the PBS had these `=infinity` lines, I think since I copied your template from the ansible docs, but I removed them since they were leading to the disk running out of space (e.g. due to systemd log files or service's memory usage running out of control). Maybe we want to remove them here too?

deploy.sh
```diff
@@ -0,0 +1,130 @@
...
+  cat >>"$BE_DEST/.env" <<EOF
+SERVER_DATA_ROOT=$DATA_DEST
+FASTIFY_PORT=$PROXY_PORT
+#   root address?
+EOF
+  
+)
+
+# FIXME this needs perms! And root user ownership. Delegate to caller for now.
+false && {
```
rogup, line 116: I delegated this stuff to the caller in the PBS install scripts (see the install.sh and install-remote.sh files)

apps/back-end/package.json
wu-lee: I didn't change `start` deliberately, so that the current behaviour would continue.

I don't think you're saying you changed `start:attached`? So I'm wondering what you mean...

rogup: I changed the `dev` script. But not actually that relevant for this.

deploy.sh
rogup: Removed them in this commit
https://github.com/DigitalCommons/mykomap-monolith/commit/f39cdc34543cb1cbdbc3682d1337d35f7e580e3e

Issue #6, opened 2024-09-04
[CWM] Convert backend to Typescript & implement API type safety / validation


rogup:
## Description

Following on from the investigation in https://github.com/DigitalCommons/mykomap/issues/269 , we need to have a go at implementing ts-rest + Zod on the backend.

## Acceptance Criteria

- [x] All back-end source code is in typescript ~and fully typed~ -> Nick to work on this
- [x] We have a typescript ts-rest contract file in the API shared library, that defines our REST API.
- [x] We have a script that can generate an OpenAPI spec from our ts-rest contract
  - [ ] and maybe automatically upload it to SwaggerHub as part of our CI pipeline?
- [x] We have runtime validation of our back-end API routes using Zod
- [x] Changes to the contract are immediately flagged as type errors in the FE/BE without need for a build or code generation.

lin-d-hop:
Delivery 5/9

- [ ] Separate branch to convert backend to Typescript

wu-lee:
Back-end is now typescript, as of e5b84f86450. Initially with `@ts-nocheck` but those are now removed.

The `tsconfig.json` is using `strict`, but I'm not certain how strict the typing is however. I have a feeling we got Mykomap 3.0 to be more strict than this. Will investigate in future work, however.

wu-lee:
@rogup I've just submitted a review request for the branch `ts-restify-the-backend` here: https://github.com/DigitalCommons/mykomap-monolith/pull/15

This should cover the final two points on the list above. (But not including uploading the swagger spec.)

wu-lee:
Merged a while ago, I think this is done, so closing.

PR #8, opened 2024-09-06
Add ts-rest contract

rogup:
#### What? Why?

Related to issue #6 

This PR converts @mykomap/common into a Typescript module, built with Vite, and adds a `contract.ts` to define the REST API. This contract can be comsumed by `ts-rest` in the back-end and front-end for API implementations and calls.

We also have added a ts-rest contract -> OpenAPI spec generator and generated an OpenAPI spec. I have checked it against the old OpenAPI spec, and it looks more-or-less the same with a few syntax changes, which I don't think matter in practice.

#### Checklist

  - I updated monorepo.md with the new structure of @mykomap/common
- Added UTs (N/A)
- Checked that all automated tests pass (N/A)

### DIFF
--- a/apps/back-end/src/pluginApi.ts
+++ b/apps/back-end/src/pluginApi.ts
@@ -3,7 +3,7 @@ import openapiGlue, { FastifyOpenapiGlueOptions } from "fastify-openapi-glue";
 import * as dotenv from "dotenv";
 import { Security } from "./security.js";
 import { Service, ServiceOptions } from "./service.js";
-import specification from "@mykomap/common";
+import { specification } from "@mykomap/common";
 import {
   FastifyInstance,
   FastifyPluginCallback,


rogup:
@wu-lee
Please could you review that the generated openapi spec looks fine? There are a few differences due to the way the generator library does things e.g. nesting of the 'description', but I think all info is still there and it's a valid spec that makes sense.

If it looks good, feel free to merge this into main

I'm hoping my changes to this module haven't broken the backend build after your typescript-ification

After my holiday, I can then start working on migrating to ts-rest on the FE side, and hopefully it will all connect up nicely without too much pain!

wu-lee:
Thanks, I've rebased it up to the top of the main branch, and fixed the minor conflicts with the work I've pushed prior to that.

As part of reviewing it, I also split one of the commits up ("Fix module entry point") so that the npm install and `index.ts` steps are in separate commits, mainly to highlight what those npm modules are.

The `monorepo.md` document changes are committed in an earlier branch. 

Merging this as you suggests, it seems to work fine with the back end.

All of this ends up in an identical endpoint - I've not changed the net content of the commits.

Issue #4, opened 2024-08-30
[CWM] Create API calls in FE


lin-d-hop:
** Please track in Clockify CWM project **

## Description

Based on the API design for MykoMap discussed in https://github.com/DigitalCommons/mykomap/issues/267 and the [notes from the meeting on 20/8](https://docs.google.com/document/d/1aguAEMKfr8gqZRVkhJnb46VRJKd5wPMLLOQ1UmAKgFo/) and the API docs being created in #2 let's build the API requests into the FE.

## Acceptance Criteria

- FE is fetching data for main map call and marker data call from the BE. This data is currently mocked to my understanding.
- Do we need any translations at this stage?
- Do we need any unit tests?

rogup:
I used https://github.com/hey-api/openapi-ts to quickly generate API calls from the openapi spec.

Then linked the UI up to these calls via the Redux Filter slice.

Since the full set of locations is now fetched async, I had to tweak the way this is initialised and stored too, but the architecture is roughly the same: there is a `features` variable in `geojson.ts` which holds a copy of the data that we don't change, which can be imported by other parts of the front-end app. I wanted to make this `features` object readonly using `Object.freeze()` but this didn't work with MapLibre, which doesn't want the geojson data to be readonly.

The changes are in `main`.

rogup:
This is done for some mock data https://dev.maps.coop/cwm/?datasetId=test-500000
On this demo, a dataset of 500000 markers is loaded, and you can search for 'baz', which performs a filter.

The API calls were generated from the OpenAPI spec using @hey-api/openapi-ts, for speed of development, so we could get this working skeleton. But I'm keeping this ticket open since we are now going to migrate to using ts-rest.com

rogup:
We have migrated to ts-rest so this can now be closed

PR #10, opened 2024-09-18
Use ts-rest fetch client on front-end to interact with API

rogup:
#### What? Why?

Related to issue #4 

Since we have added a `ts-rest` contract to define the api in the `common` lib, we are using a `ts-rest` client on the front-end to consume the contract and benefit from the type safety.

There are also some other slightly related changes in this PR which I ended up doing at the same time, but they are clearly separated in the commit history.

#### Checklist

- [x] Updated README
- n/a Added UTs
  - this code itself doesn't really need UTs, but we can make use of ts-rest Fake APIs later when we want to add automated regression tests for the front-end (e.g. with Playwright or Backstop JS) https://ts-rest.com/docs/guides/fake-apis
- n/a Checked that all automated tests pass


rogup:
@wu-lee Please can you give these changes a quick review? (it's easier to have a look at each commit in order)

wu-lee:
Seem to make sense, no glaring questions. I'm not going to sign off every commit as that seems excessive, just offer a general :+1: 

Some prettification mixed in with changes mean I need to mentally skip those bits. 

Some changes seem unrelated to the branch topic, like the GH pages removal?

Aside: I find that `package-log.json` to be a source of major meaningless churn and so I tend to put changes to this file in their own commits, at the end of a branch after everything has settled down. Not a big deal, but it does help when rebasing around other people's commits when they also change that file - as you tend to have to stop and resolve the conflicts every time it gets touched, typically by deleting it and running `npm install` again.

rogup:
@wu-lee
Thanks.

I can't see any prettification happening in these commits.

Yes, there are a couple of commits which are a bit unrelated, which I mentioned in the description of this PR. I should probably have stashed them for later, but I think it's fine since I won't be doing a squash merge anyway, and they're in separate commits.

Good point about the `package-lock.json`, I'll do that in one commit in the future to help with rebasing.

PR #15, opened 2024-10-04
Ts restify the backend

wu-lee:
#### What? Why?

Related to issue #6

This PR converts the back end from using the `fastify-openapi-glue` implementation to an equivalent one using the new `ts-rest` based contract. It also fixes the test suite (such that there is) and adjusts the contract to match the API in use, in particular with respect to the errors (they now return messages), the dataset ID (it's now a number) and the version endpoint (the response is more specific, not just a pass-through).

#### What should we test?
- Check it builds (`dev` and `build` front and back end) and that intellisense works in your editor.
  - The build should also now update the `common` package to ensure the contract is up to date
- Check the version endpoint works and provides sensible information. You may want to add some test tags to your working repo to see how that changes the version information.
- Check the tests pass.

#### Checklist

- [ ] ~~Updated or added any necessary docs in `docs/`~~ I dont think the documentation there needs additions? I have however included TSDoc style comments to the new code I've added.
- [ ] ~~Added UTs~~ (Not applicable)


apps/back-end/src/routes.ts
```diff
@@ -0,0 +1,189 @@
...
+type Contract = typeof contract;
+
+//////////////////////////////////////////////////////////////////////
+// Helper functions
+
+// Helper functions
+
+/** A predicate testing for URL-safe base64 string (RFC4648 sect 5) */
+function validBase64(value: string): boolean {
+  return !!value.match(/^[A-Z0-9_-]+$/i);
```
rogup, line 30: If I understand the docs correctly, ts-rest should automatically do runtime validation of the request and return 400 if it fails, so we don't need to do it separately
https://ts-rest.com/docs/fastify#request-validation-error-handling

The Zod schema in `contract.ts` just needs the datasetID and filters to be strings, so we maybe need to tighten up these constraints in the schema, then we can get rid of these functions.

apps/back-end/src/routes.ts
```diff
@@ -0,0 +1,189 @@
...
+      params: { datasetId, datasetItemId },
+      request,
+      reply,
+    }) {
+      if (!validBase64(datasetId))
+        throw new TsRestResponseError(contract.getDatasetItem, {
+          status: 401,
+          body: { message: `invalid datasetId: '${datasetId}'` },
+        });
+      if (datasetItemId < 0)
```
rogup, line 161: This can probably be added to the Zod schema in `contract.ts` too.

apps/back-end/test/data/datasets/some-file/items/0.json
```diff
@@ -0,0 +1 @@
+{"id":0, "a":"a:foo", "b":"b:bar", "desc": "humbug"}
```
rogup, line 1: Whilst we're changing these file names, I'm thinking the folder name `some-file` could also be changed to something like `dummy-dataset` to make the test data more instructive as an example.

apps/back-end/src/app.d.ts
```diff
@@ -0,0 +1,2 @@
+// Tell TS what this is
+declare const __BUILD_INFO__: string;
```
rogup, line 2: This is more of a question- why does this get defined as a string, when it seems to be defined as an object in the Vite build config?

rogup (commented):
Looks good, have left a few comments

apps/back-end/src/routes.ts
wu-lee: Main reason for not doing that is I just don't currently know how to make Zod perform arbitrary validations on strings.

apps/back-end/src/app.d.ts
wu-lee: Ah, good catch - that's a mistake. Although one you'd think the TS validation would catch, in the build or in the editor...  does it do that for you?

apps/back-end/test/data/datasets/some-file/items/0.json
wu-lee: I can change that if you like.

apps/back-end/src/routes.ts
rogup: There's a `z.string().regex()` validation and also a built-in `z.string().base64()`

wu-lee: I've addressed this general point in some fixup commits (labelled with `fixup`) later in the chain: all this sort of validation which can be is now done via the contract.

Let me know if you want me to leave those separate before you review them, because I intend to merge the fixup commits with the commit they are modifying before merging the PR. You might prefer to review after I've done that?

apps/back-end/src/routes.ts
wu-lee: Ditto above.

apps/back-end/src/app.d.ts
wu-lee: Resolved in a fixup commit.

wu-lee:
This whole branch has been rebased to the end of `main` and some minor conflicts resolved (just package-lock.json)

apps/back-end/src/routes.ts
rogup: Looks good, feel free to merge the commits however makes sense

apps/back-end/test/data/datasets/some-file/items/0.json
rogup: I'm now seeing no test data in https://github.com/DigitalCommons/mykomap-monolith/tree/ts-restify-the-backend/apps/back-end/test

apps/back-end/src/routes.ts
```diff
@@ -0,0 +1,147 @@
...
+          body: { message: `item retrieve failed` },
+        });
+
+      return reply;
+    },
+
+    async getVersion(req) {
+      return {
+        body: __BUILD_INFO__,
+        status: 201,
```
rogup, line 143: Should this be 200?

wu-lee: I think this is a hang-over from the swagger-generated version wanting to use 201 for successful responses ("created successfully") - in this case (and the other cases in the API)  nothing is created, so I agree 200 is better.

apps/back-end/test/data/datasets/some-file/items/0.json
wu-lee: The test data *should* still be in `apps/back-end/test/data/datasets`. But I see it isn't... investigating and fixing.

wu-lee: Ok, just an accident - the new versions weren't staged in the commit. Fixed in an appended fixup commit for later integration.

wu-lee:
Fixes for above changes requested now added as fixup commits above, then the branch rebased on to the current `main` branch to verify it can be integrated.

Then fixups squashed into relevant commits, or promoted to primary commits, as appropriate.

Result rebased again to integrate a fix added to main branch.

Issue #20, opened 2024-10-10
[CWM] Create the functional back-end


rogup:
## Description

Create the functional back-end based on the design #14 https://docs.google.com/document/d/1Y2MZ-Q92bhFCit-SoivpriOsxQZIbJop4ewiCqkYRNU/edit

## Subtasks

It probably makes sense to do each of these tasks in its own PR. Write UTs for all the PR's changes within the same PR, rather than leaving them to the end. Try using [BDD](https://github.com/DigitalCommons/land-explorer-front-end/wiki/Automated-Testing) (maybe copy that LX wiki to our codebase docs)

- [x] create the overall app structure, including the classes (e.g. Dataset) and placeholder class instance methods, based on the architecture diagram. #23 
- [x] create simple mock files in the `test/data` directory, based on the file structure in the design and confirmation of what categories we need from Marcel/Colm #24 
- [x] use these mock files to write UTs for the API routes that have already been implemented (i.e. the ones that simply read and return a JSON file) #52 
- [x] implement methods to Dataset class which serve the locations.json, vocabs.json, and individual initiatives <id>.json files #52
- [x] implement method to Dataset class to perform a search using the searchable.json #52

## Acceptance Criteria

- [x] A functional back-end that matches the design
- [x] Full UT coverage of the back-end logic that runs in GitHub CI
- [x] Codebase docs to describe the architecture and any other important details

wu-lee:
Seems mostly okay, I've added some review comments added for PR #23, some of which we already discussed before on a call but I'm adding for the record (not sure of our exact conclusions).

ColmDC:
Assuming this is closable.

PR #23, opened 2024-10-14
[CWM] Create back-end class structure

rogup:
#### What? Why?

This PR is for task 1 of #20 

I've created the `datasetService` and class structure, ready for the remaining methods in the Dataset object to be implemented.

Note that this new `datasetService` isn't hooked up the API routes yet, since the new file structure isn't implemented in the `test/data` so some tests would fail. It made sense to wait after this mock data has been created (we're waiting on the spec from Colm and Marcel for this) before hooking them up, removing any unused implementation code from the old skeleton server, and adding UTs.

#### Checklist

- Can't add UTs yet, this is task 3 of #20 


rogup:
@wu-lee Please can you review these changes?

docs/architecture.md
```diff
@@ -18,3 +18,94 @@
...
+          "sz:large": "Large",
+          "sz:medium": "Medium",
+          "sz:small": "Small"
+        }
+      }
+    },
+    ...
+  },
+  "popups": { "field": "sz", "style": "text" }, { "field": "websites", "style": "clickable-list" },
+  "ui": { ... },
```
wu-lee, line 87: Not sure what this is?

docs/architecture.md
```diff
@@ -18,3 +18,94 @@
...
+`config.json`:
+
+```
+{
+  "prefixes": {
+    "https://example.com/sizes/1.1/": "sz",
+    ...
+  },
+  "vocabs": {
+    "sz:": {
```
wu-lee, line 74: I think we agreed to drop the trailing colon here?

docs/architecture.md
```diff
@@ -18,3 +18,94 @@
...
+      "EN": {
+        "title": "Sizes",
+        "terms": {
+          "sz:large": "Large",
```
wu-lee, line 78: I think we agreed to drop the `sz:` prefix here?

docs/architecture.md
```diff
@@ -18,3 +18,94 @@
...
+
+### Data generation
+
+These directories of JSONs, including the searchable strings in the `searchable.json` files, need to be pre-generated by a script. This script will be written in JS/TS and live in the monorepo, to be run on the back-end server.
+
+The script will take the full data CSV for a map (generated by the data factory) as inputs, and write the full data into the required JSON files in the directory structure specified above.
+
+#### Note:
+
+We will need to manually copy the `standard.csv` from the data factory server to the back-end. Maybe in the future, the data factory pipeline can be enhanced to write the JSON files to the back-end server so that no manual duplication is necessary (and maybe we can eventually get rid of the separate data server altogether).
```
wu-lee, line 101: Or, perhaps better, the mykomap server could be given an URL to the appropriate standard.csv file(s) as published by the data factory and download it from there (possibly when notified by a webhook, or possibly polling and checking the file modification date)

docs/architecture.md
```diff
@@ -18,3 +18,94 @@
...
+
+### Dataset instances
+
+- For each dataset available in the `datasets` directory on server start, a dataset instance is created
+  by the Dataset service. Each Dataset instance has a:
+  - `searchable` property, which is just the `searchable.json` loaded as an in-memory object
+  - `getItem` method
+  - `getConfig` method, which includes the vocabs
```
wu-lee, line 109: This could be be streamed by the back-end too, in which case it would need a path as below, or a method which returns a stream.

docs/images/architecture-front-end.drawio.svg
wu-lee, line 1: Can't see any differences here?

docs/architecture.md
```diff
@@ -18,3 +18,94 @@
...
+`locations.json`:
+
+```
+[ [1.21419, 50.45254], [0.21002, 49.33954], … ]
+```
+
+`searchable.json`:
+
+```
+[ { "coun": "coun:GB", "sz": "sz:small", "searchString": "some co-op 2 green lane london n4 9qr" }, … ]
```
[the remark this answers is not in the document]
wu-lee: For related reasons I'm wondering if the really bulky bit, the text `searchString` part, should be kept in its own file. This file could just be a plain text file, with one line per item. Searching it could be done by streaming it from disk, which avoids loading the entire file permanently into memory (for each dataset).

For instance, this SO [thread][1], top of the google search I just made for "javascript search text file streaming", has some sample stream-searching code, and a reference to a module which performs the streaming by what appears to be a fast non-buffering algorithm.

[1]: https://stackoverflow.com/questions/20187145/how-to-search-stream-for-string-in-node-js
[2]: https://github.com/mscdex/streamsearch

docs/architecture.md
rogup: Yet to be defined properly, but I think UI stuff like htmlTitle, aboutHtml, showDatasetsPanel that used to exist in the old ConfigData

docs/architecture.md
rogup: Added to the docs

docs/architecture.md
```diff
@@ -18,3 +18,94 @@
...
+We will need to manually copy the `standard.csv` from the data factory server to the back-end. Maybe in the future, the data factory pipeline can be enhanced to write the JSON files to the back-end server so that no manual duplication is necessary (and maybe we can eventually get rid of the separate data server altogether).
+
+### Dataset instances
+
+- For each dataset available in the `datasets` directory on server start, a dataset instance is created
+  by the Dataset service. Each Dataset instance has a:
+  - `searchable` property, which is just the `searchable.json` loaded as an in-memory object
+  - `getItem` method
+  - `getConfig` method, which includes the vocabs
+  - `getLocationsFilePath` method (which is then streamed to the Fastify response)
```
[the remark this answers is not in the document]
rogup: Ah yeah, that's more elegant. I was trying to work out how to stream from the Dataset class without it having to know about Fastify

docs/architecture.md
rogup: I'm not sure if streaming works with runtime Zod response validation, I'm guessing not. So I thought it made sense to not stream in this case, unless we expect the config files to get very large. Could always change it in the future if needed

docs/images/architecture-front-end.drawio.svg
rogup: It was a slight difference to rounding of an edge, since I needed to copy and paste the rectangle

docs/architecture.md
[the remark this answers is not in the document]
rogup: I think these are optimisations that we don't need to make yet for the number of markers and filterable fields in the Delhi demo. The server will have enough memory. We need to prioritise getting a back-end working and not spend on time on nice-to-haves. They will add more complexity to the code rather vs just reading in the JSON straight into a `searchable` object.

I will note these optimisations in the docs for reference - the code isn't set in stone and it will be easy to make them in the future if needed.

docs/architecture.md
```diff
@@ -18,3 +18,94 @@
...
+│   │   ├── initiatives
+│   │   |   ├── 0.json (full info of first initiative in the above aggregate JSONs)
+│   │   |   ├── 1.json
+│   │   |   ├── ...
+│   ├── other-dataset
+│   │   ├── ...
+│   ├── ...
+```
+
+Additionally, for each map variant, there's a `config.json`, which contains config for displaying the map in the UI, including the vocabs (translations of data IDs), default sidebar panel, and popup appearance.
```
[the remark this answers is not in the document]
rogup: I thought `config` would also code UI related things like the old `config.ts`, and would need to be checked into the source code, rather than generated by a script like the actual data. But I think it's maybe fine to have it per-dataset rather than per-map. You're right though, these docs imply configs aren't associated with a specific dataset. Maybe they don't need to be if we want to share datasets in the future but I think it over-complicates things for now, so it's better to have one config per dataset and stick with linking them to a datasetID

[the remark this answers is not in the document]
rogup: I've updated these docs to reflect my above comment, which isn't very clear. And I've also updated the folder structure in PR #24 to match this

docs/architecture.md
wu-lee: Ok.

docs/architecture.md
wu-lee: Yes, streaming bypasses ts-rest's on-the-fly Zod validation, I think. However, this is not to say that such streamed files couldn't be validated on creation, and I think they should be. Do we benefit by validating every request after that? Not sure, suspect not much.

But overall this is just a suggestion, as you say the configs will be relatively small.

docs/architecture.md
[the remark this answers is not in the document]
wu-lee: I think the datasets (and configs) could be version controlled, and in fact I have been doing that - I just checked the whole datastore into git. Would we want the config to be version controlled separately? Not sure, I think it might not need to be though.

PR #24, opened 2024-10-17
[back-end] Add test data

rogup:
#### What? Why?

Related to issue #20 

In our delivery meeting, we confirmed that this mock data covers all the fields we need in the Delhi demo
```
{
  "Identifier": "test/cuk/R000001",
  "Name": "Test Co-op",
  "Description": "We sell some stuff",
  "Website": "https://test.coop",
  "DC Domains": "test.coop;testlong.coop",
  "Country ID": "GB",
  "Primary Activity": "ICA230",
  "Organisational Structure": "OS60",
  "Typology": "BMT30",
  "Latitude": 51.507476,
  "Longitude": -0.127825,
  "Geo Container Latitude": 51.50747643,
  "Geo Container Longitude": -0.12782464,
  "Geocoded Address": "1 West Street, Sheffield, S1 2AB, UK",
  "NCBA Member": 0,
  "ICA Member": 0,
  "DC Registered": 1,
  "CUK Member": 1,
  "CUK Sector": "Water"
}
```

We need to add some demo JSON files based on this, so we can start adding UTs.

#### Checklist

- [ ] Added UTs


wu-lee:
I think multiple-valued items like this:

    "DC Domains": "test.coop;testlong.coop",

Would be better as arrays?

    "DC Domains": ["test.coop","testlong.coop"],

Using delimiters can be a bit fraught in general, as you need to either pick one which cannot legitimately appear in the data or implement an escaping rule. 

Necessary sometimes with CSV, but I think this can be entirely avoided in JSON.

wu-lee:
One thought about the "member of" fields. Perhaps these would  need to be (or more conveniently be) collapsed into a single field, perhaps an array or an object.

Whether this is true or not depends a bit on how we handle them in the filters - the UI design I've seen wants them to be one filter, not N filters.

rogup:
> I think multiple-valued items like this:
> 
> ```
> "DC Domains": "test.coop;testlong.coop",
> ```
> 
> Would be better as arrays?
> 
> ```
> "DC Domains": ["test.coop","testlong.coop"],
> ```
> 
> Using delimiters can be a bit fraught in general, as you need to either pick one which cannot legitimately appear in the data or implement an escaping rule.
> 
> Necessary sometimes with CSV, but I think this can be entirely avoided in JSON.

Updated this

rogup:
> One thought about the "member of" fields. Perhaps these would need to be (or more conveniently be) collapsed into a single field, perhaps an array or an object.
> 
> Whether this is true or not depends a bit on how we handle them in the filters - the UI design I've seen wants them to be one filter, not N filters.

Yes, good idea. I think this makes more sense. I've updated the test data

wu-lee:
All good for merging I think.

Issue #19, opened 2024-10-09
[CWM] Add vocabs API route to the ts-rest contract 


wu-lee:
## Description

The mykomap client will need to get information about pin attribute categories.
Specifically:
- What categories exist. For each:
  - A unique category ID
  - A list of term IDs
  - For each language supported by the map (although this might be only one language at a time):
    - A title for the category as a whole
    - A title for each term

This information was provided by the `vocabs.json` format in older versions of Mykomap. It might be convenient to retain that format/structure here.

Examples of this can be visible on the data server. Some specific examples:
- https://dev.data.solidarityeconomy.coop/ncba/vocabs.json
- https://dev.maps.coop/demo-merge-map-plus/ncba-vocabs.json

A cut-down example from the Mykomap documentation follows. This defines one vocabulary defining standard size definitions. The SKOS vocab URI for this is `https://example.com/sizes/1.1/`.  An abbreviation prefix is defined for use in this context, `sz`. Within the rest of the JSON document, a `sz:` prefix represents that URI, and the URI for a term can be reconstructed by replacing the prefix and the colon with that URI.

Languages are identified by a two-character (upper case) code as defined by [ISO 639 Set 1][iso639].  There is only one language section here, for `EN`, with term labels in English. But there could be as many as required.

Both the prefixes and the language codes should be unique within this file. The same set of languages should be present for all vocabs. (When they are not, this might be an error, or it might be resolved by rules to fall-back to some other language, but that's outside the scope of this proposal.)

[iso639]: https://en.wikipedia.org/wiki/List_of_ISO_639_language_codes

```
{
  "prefixes": {
    "https://example.com/sizes/1.1/": "sz"
  },
  "vocabs": {
    "sz:": {
      "EN": {
        "title": "Sizes",
        "terms": {
          "sz:large": "Large",
          "sz:medium": "Medium",
          "sz:small": "Small"
        }
      }
    }
  }
}
```
### Use as language files

The older version of Mykomap was also starting to use these vocabs for language translations, which are functionally very similar. 

Not all translations can (or should) be hardcoded into the application, either the old or the new one:
- Specific text for custom layouts may be required. 
- Languages not included in the base application may need to be supported.

See the `ui:` vocab in the second example above for an example of a vocab used to define the labels for UI text not built into the Mykomap application.

### Linked data interoperability

These vocab files were designed to be interoperable with linked-data SKOS vocabularies (hence the name "vocab" rather than "taxonomy" or "category")

[...]

PR #32, opened 2024-10-21
Add vocabs route

rogup:
#### What? Why?

Related to issue #19 

#### What should we test?

- Go to '...'
- 

#### Checklist

- [ ] Added UTs


rogup:
@wu-lee I've created this PR so that I can view the changes more easily, please can you fill in the details when you get a chance

libs/common/src/api/contract.ts
```diff
@@ -1,73 +1,55 @@
...
- *
- * FIXME Perhaps we should also disallow common URI scheme prefixes.
- * FIXME Maybe we don't care about unicode?
- *
- * This regex requires node 10+ to be able to use /u and \p.
- *
- */
-const QNameRx = new RegExp(`^${NCName}[:]${NCName}$`, "gsu");
+function ZodRegex(rx: RegExp, message: string) {
+  return z.string().refine((v: any) => RxUtils.uaon(rx).test(String(v)), {
```
rogup, line 22: @wu-lee What's the reason for using `.refine` vs `.regex` since the former looks more complicated and verbose? This is a question rather than a piece of feedback, I'm new to Zod

rogup: Hmm I think I might understand this... it's so that the regex can be built up programatically rather than statically hardcoded as regex string?

I feel like pre-building outside of the code then hardcoding the regex, and simply commenting what it means, would be less confusing for readers. But I suppose it makes it harder to tweak the regex later, since the work of building it would need to be redone from scratch, so this way makes sense

wu-lee: > @wu-lee What's the reason for using `.refine` vs `.regex` since the former looks more complicated and verbose? This is a question rather than a piece of feedback, I'm new to Zod

I was using `.regex()` but adopted `.refine()` from somewhere I found researching Zod - I forget where, perhaps an issue or the documentation (which does have usage examples I think I used).

The main advantage I can see here is that the error message on failure states the reason more clearly - i.e. I think you get "This is not a valid QName" vs "This does not match the regular expression <insert noise here>".

Another advantage here is that it does some wrapping of the pattern: it adds anchors to it. This is so that the regular expressions defined in the code can all be kept unanchored patterns. Which means they can be re-used as components. For instance, Rx.NCName is used to build Rx.QName, which it couldn't be if it were anchored to match the whole string. Likewise Rx.QName is used in Rx.PrefixURI's definition.

wu-lee: > Hmm I think I might understand this... it's so that the regex can be built up programatically rather than statically hardcoded as regex string?

Yes. And as I say, the bits are then reusable, and NCName can be used to define QName, which can be used to define PrefixURI.
 
> I feel like pre-building outside of the code then hardcoding the regex, and simply commenting what it means, would be less confusing for readers. But I suppose it makes it harder to tweak the regex later, since the work of building it would need to be redone from scratch, so this way makes sense

I started doing that, then realised how much duplication there was. The regexes got unwieldy quickly.

rogup: Thanks, this makes sense

Issue #26, opened 2024-10-17
[CWM] rename getDataset to getLocations


wu-lee:
## Description

Suggested in #19:
              @wu-lee When you make these changes, I'm thinking we should also change the `getDataset` route to `getLocations` since it only returns the locations

_Originally posted by @rogup in https://github.com/DigitalCommons/mykomap-monolith/issues/19#issuecomment-2411561388_

wu-lee:
PR #34 added to address this.

wu-lee:
PR #34 now merged so I think this is now addressed.

PR #34, opened 2024-10-24
Rename getDataset

wu-lee:
#### What? Why?

Related to issue #26

The name `getDataset` suggests something which doesn't happen.

I've not followed the suggestion exaxctly, see the comment in the first commit.

#### What should we test?
Test all the tests run ok and the app still nominally works when run.

#### Checklist

- [ ] Added UTs

#### Deployment notes

N/A


wu-lee:
Note this branch is currently tacked onto the end of another branch under review. Will need to be rebased back onto main later.

rogup (approved):
Looks good!

wu-lee:
Having answered @rogup question directly, the reviewer seems to be happy for this to be merged, which is now done. (See above for the merge commit, which references this PR. There was some rebasing to do since it was chained onto another branch for operational reasons, hence the merge done in git outside of GitHub, which makes it think this was closed rather than merged.)

Issue #25, opened 2024-10-17
[CWM] rename references to "identifiers" to "indexes" in the API where approprate


wu-lee:
## Description

In the API, when referring to an item in a dataset we use the term `datasetItemId`. But in fact it's not an identifier, it's an index. We should use the correct name to avoid confusion, so perhaps `datasetItemIx`. 

It is possible however that we would still want to get items by their ID, as these are persistent through re-ordering of the items, versions of the data, and often across different datasets. So perhaps we need a second API endpoint, or perhaps one endpoint with a way to disambiguate which kind of reference we're using.

## Acceptance Criteria

- [x] we have an API endpoint which allows accessing items by identifier
- [x] we have an API endpoint which allows accessing items by index
- [x] the naming is consistent and sensible
- [x] application works and tests all pass

wu-lee:
I've added a PR #33 with a proposed implementation.

wu-lee:
PR #35 merged following review.

I've checked the boxes above, but maybe I should check with you @rogup that you think this is all addressed?  If so please go ahead and close this.

PR #35, opened 2024-10-24
Allow Indexes and IDs in getDatasetItem API endpoint

wu-lee:
#### What? Why?

Related to issue #25

In essence, instead of having two endpoints, one for getting items by their index, and one for getting them by ID, just adapt the existing one.
- Allow the getDatasetItem endpoint, which has the route `/dataset/:id/item`, to accept indexes and ids
- But in a way which it can be determined which is which
- Return the appropriate item, obtained via the index or the ID, in the usual format as before.

#### What should we test?
Check the reasoning makes sense, and the tests test what they are meant to.

#### Checklist

- [x] ~~Updated or added any necessary docs in `docs/`~~
- [x] Added UTs

#### Deployment notes

None


wu-lee:
Note, this branch is also appended to some prior branches so the PR is not nearly as long as it seems!  These branches will all be merged separately.

rogup:
Hi @wu-lee, I found it difficult to review this due to the PR's changes being mixed in with lots of other commits (I usually do reviews directly on GitHub, unless I need to do some checks in my IDE).
But it looks generally fine, and I'm happy for it to be merged so that stuff can be progressed, and we can fix anything that comes up later.
I just have 1 comment about decoding the IDs/indexes

wu-lee:
Having answered @rogup question directly, the reviewer seems to be happy for this to be merged, which is now done.

Issue #29, opened 2024-10-21
[CWM] Link front-end filter panels to working search/filter methods


lin-d-hop:
## Description

Following the completion of [#272](https://github.com/DigitalCommons/mykomap/issues/272) we need to connect together the FE filter panels to the functional search/filter methods.

I would suggest a first step will be for @ms0ur1s and @rogup to have a call and break down this epic into a checklist of tasks. That process looks like it works really well. Nice one team :raised_hands: 

## Acceptance Criteria

- All panel functionality is working as expected
- I18N should be in place, though the language switcher itself is out of scope for this issue
- Checklist of tasks is fully completed

ColmDC:
Assuming this should have been closed a while ago.

PR #40, opened 2024-11-04
29 Link Redux to search UI and add mock config

rogup:
#### What? Why?

Related to issue #29 

This PR links up the new search panel to Redux and adds some mock config to populate the panel.

The directory panel UI needs to be fixed to use buttons rather than links before it can be hooked up to Redux.

Note that the labels for the filters are still just the field IDs, logic needs to be added for the translation of these IDs into text.

#### What should we test?

- Open the search panel and see the drop-down for each filter option.
- Change language to French by adding `lang=fr` to the URL query params and check that the filter options are now in french.

#### Checklist

- [x] Added UTs


rogup:
@ms0ur1s Please could you review the things that I've added to this branch since the commit that you added? Please let me know if there's anything that you'd like explaining, and once you're happy with the changes, we can merge this in main and deploy it for the demo.

rogup:
I will also add some more UTs tomorrow, since we should add some tests for the Redux selector logic in searchSlice

rogup:
@ms0ur1s since you're away for a few days, I've merged this so we can get Demo 1 out, but please still give this a review and I can make changes in a separate PR

ms0ur1s:
@rogup, will do man. I'll give it a look before I merge create a branch for the MUI component switch

ms0ur1s:
@rogup, when I try to run the back-end I receive the below error, what am I doing wrong.
image

ms0ur1s:
As per the above I can't get the back-end to run at the moment (most likely my own fault), but there a few UI idiosyncrasies that I spotted so I'll make a quick branch to fix these. I'll add details and the branch name below.

ms0ur1s:
@rogup, although I don't have the back-end running, the updates are looking amazing. Great work! It's so good to see all the piece's starting to connect.

wu-lee:
> @rogup, when I try to run the back-end I receive the below error

That command works for me... Looking at your screenshot I see the version of node reported as 18.18.0 - from this and the error, I think you likely need to be using the one specified in `.tool-versions` - 22.4.1. Mentioned in the docs here but maybe not well highlighted: https://github.com/DigitalCommons/mykomap-monolith/blob/main/docs/deployment.md

To get this on Linux, best use ASDF, instructions here. You also need the nodejs plugin for ASDF. This may require re-starting your terminal to work.

https://asdf-vm.com/guide/getting-started.html

Finally you need to:

    asdf install nodejs 22.4.1

ms0ur1s:
> > @rogup, when I try to run the back-end I receive the below error
> 
> That command works for me... Looking at your screenshot I see the version of node reported as 18.18.0 - from this and the error, I think you likely need to be using the one specified in `.tool-versions` - 22.4.1. Mentioned in the docs here but maybe not well highlighted: https://github.com/DigitalCommons/mykomap-monolith/blob/main/docs/deployment.md
> 
> To get this on Linux, best use ASDF, instructions here. You also need the nodejs plugin for ASDF. This may require re-starting your terminal to work.
> 
> https://asdf-vm.com/guide/getting-started.html
> 
> Finally you need to:
> 
> ```
> asdf install nodejs 22.4.1
> ```

Thanks @wu-lee, I'll give this a try.

Issue #30, opened 2024-10-21
[CWM] Write a script(s) to import datasets


lin-d-hop:
## Description

Data is incoming for the CWM. [This spreadsheet](https://docs.google.com/spreadsheets/d/1bIVGZ1O56NVURZ9Zw7yDdz2sALTKEWgVxhYwY3hwwfM/edit?gid=0#gid=0) lists the data sets that have been shared, and some that are expected.
We need to import this data.

The initial data imports will be:

- [x] NCBA Members
- [x] DotCoop registrants(.coop & marque)
- [x] ICA
- [ ] Co-ops UK Open Data
- [x] USDA
- [x] NCG
- [x] US Federally Insured Credit Unions
- [x] Farm Credit Administrators
- [ ] IFFCO membership data
- [ ] Cooperatives Mutuals Canada Federation
- [ ] US Worker Co-op Federation

All of these are linked from the spreadsheet. Other data sets are expected to be incoming while this issue is in play, but I think it would be prudent to scope the work at these datasets and extend the scope if resources allow. I want to make sure we have capacity for bug fixing once we get to start doing more thorough end to end testing.

## Acceptance Criteria

- The above data sets have been imported and are visible on our megamap.

ColmDC:
Added 2 sets to the list in the description. This is final list of data sets for Delhi as 25th Oct was cut off date.

ColmDC:
This ticket doesn't mention merging. Shall we treat that as a second ticket?

wu-lee:
*Step 1*. I think  the entry-point to our systems is via the data factory - every data set needs something to normalise it and geocode the addresses, which is what the DF does.  The results get published on the web as `standard.csv` files (whether or not that's good name), plus other bits (like a vocabs.json file, and lots of TTL and RDF files). That could be part of this issue?  But is not described explicitly yet.

Then there's a *"Step 2"* which is the merging. You're right, this might be a separate thing, so might warrant another issue.  The output of this is probably a table or organisations, (maybe a CSV, akin to the standard.csv above), plus some category indexes used by that table (maybe a vocabs.json fille). This will get big. We may want to think about keeping track of both the inputs and the output via something version-controlly, but perhaps not conventional version control as that's very focussed on text rather than data.

Then there's the script which takes that raw data and munges it into whatever mykomap wants on the back end.  Which is *Step 3* and what this issue seems to concern.

[...]

PR #51, opened 2024-11-08
add PropDefs to config

wu-lee:
#### What? Why?

Related to issue #30

The PropDefs are interfaces used in Mykomap 3.x to describe the properties (fields) of the data being mapped. We need a similar description to be able to manipulate data for the import script, and for the back-end search/filter implementation.

This PR just covers these definitions, plus logic for creating and manipulating them using equivalent classes, which are new.  The classes are called PropDefs, the old interfaces are renamed as PropSpecs.  Most of the ideas and logic remains the same, but the classes package the logic in a more convenient way and avoids the need for switch statements to handle the discriminator field. 

Corresponding Zod validators are added for inclusion of the PropSpecs in the ConfigData tupe, since inferring them from types seems (from research) not to be a feature which currently exists anywhere in a robustly implemented form.  This therefore means that configs now need a `itemProps` field, but that may be an empty object.

The ConfigData structure will be used to guide the imports of CSV data in an upcoming PR. This needs to allow the fields to be imported to vary depending on the CSV file, which will have a similar but varying set of fields depending on the source.

#### What should we test?

Tests should still pass. No new tests added here.
 
Zod validators have been tested in later commits to allow assignment of the result to PropSpec types.

#### Checklist

- N/A: Updated or added any necessary docs in `docs/`
- N/A: Added UTs - will be added later in import script PR

### DIFF
--- a/apps/front-end/src/components/panel/searchPanel/searchSlice.ts
+++ b/apps/front-end/src/components/panel/searchPanel/searchSlice.ts
@@ -91,13 +91,13 @@ export const searchSlice = createAppSlice({
   extraReducers: (builder) => {
     builder.addCase(configLoaded, (state, action) => {
       const config = action.payload;
-      state.filterableFields = config.ui.filterableFields.map((field) => {
-        const fieldDef = config.fields[field] as VocabPropDef;
+      state.filterableFields = config.ui.filterableFields.map((name) => {
+        const propDef = config.itemProps[name] as VocabPropDef;
         return {
-          id: field,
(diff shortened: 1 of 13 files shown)


wu-lee:
Some fixes for broken include paths integrated. Minor alteration to front-end added to get it to compile, see relevant commit.

Tests pass, although there are warnings and what appear to be errors printed in the front-end tests.

rogup (approved):
This all looks great, I'm merging it :)

PR #53, opened 2024-11-11
30 add prop defs to config part 2

wu-lee:
#### What? Why?

Related to issue #30

Some further alterations necessary for the upcoming final PR for issue #30 which implements the dataset importer.
- The Location Zod validator must permit null locations
- Export the inferred types from the contracts, so they can be used.
- some properties (`strict`, `as`) were not captured by the PropDef constructor
- a jsdoc clarification and openapi regeneration

#### What should we test?

Tests should all pass as before. 

#### Checklist

- N/A] Updated or added any necessary docs in `docs/`
- N/A Added UTs

### DIFF
--- a/libs/common/src/api/contract.ts
+++ b/libs/common/src/api/contract.ts
@@ -27,7 +27,7 @@ function ZodRegex(rx: RegExp, message: string) {
 const Location = z.array(z.number()).min(2).max(2);
 const DatasetId = z.string().regex(Rx.UrlSafeBase64);
 const DatasetItem = z.object({}).passthrough();
-const Dataset = z.array(Location);
+const Dataset = z.array(Location.nullable());
 const NCName = ZodRegex(Rx.NCName, "Invalid NCName format");
 const QName = ZodRegex(Rx.QName, "Invalid QName format");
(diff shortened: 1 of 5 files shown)


libs/common/src/api/contract.ts
```diff
@@ -27,7 +27,7 @@ function ZodRegex(rx: RegExp, message: string) {
 const Location = z.array(z.number()).min(2).max(2);
 const DatasetId = z.string().regex(Rx.UrlSafeBase64);
 const DatasetItem = z.object({}).passthrough();
-const Dataset = z.array(Location);
+const Dataset = z.array(Location.nullable());
```
rogup, line 30: sorry, pedantic request but think this is the best time to do it: please can you rename this type to DatasetLocations to be a bit clearer?

rogup (changes requested):
Everything looks good to merge after the small change I requested :)

libs/common/src/api/contract.ts
wu-lee: Ok, 2 commits for this added to the end of the PR.

rogup: Looks good

PR #60, opened 2024-11-19
29 Get AboutPanel working

rogup:
#### What? Why?

Related to issue #29 

This PR adds support for About markdown text to be server from the back-end, and fixes up the front-end to correctly display this.

@ms0ur1s Will hand this PR over to you to review and maybe do more work on. Think most of it is done, although not sure if a story is set up for it (and I think a lot of the stories are actually broken in general, since I don't know how they work, and they're not included in the automated tests. It would be good to do a skill-sharing for this after this week and write up a process for writing them in a doc)

#### Checklist

- [x] Added UTs

- Open about panel and check is looks ok.
- Does this need to support translations @ColmDC ? If so, this work will need to be extended


rogup:
Ok I'll just merge this now so it can be QA'ed

rogup:
@ColmDC / @lin-d-hop
I've added this about text for the Delhi demo, feel free to edit that file directly in GitHub in the delhi-demo branch if you'd like to change it, then we can git pull it onto the server

https://github.com/DigitalCommons/cwm-test-data/blob/delhi-demo/datasets/delhi/about.md

Issue #66, opened 2024-11-20
[CWM]  Critical mobile UX bugs


lin-d-hop:
## Descriptions

In initial mobile testing a number of critical UX issues have come up. These all need fixing.

- [x] Directory, Search and results don't scroll up quite far enough, leaving some content behind the nav
- [x] Pop-up doesn't fit on the screen well, resulting in weird behaviour and a difficulty to view it, particularly on smaller screens
- [x] Add apply filter button and don't open the results panel until this is clicked. The UX on this needs to be clear ie dont make it so you ahve to scroll to see the button.
- [x] Search results not closing when result is clicked (doing this in #64)

## Acceptance Criteria

The above issues are fixed and deployed.

ms0ur1s:
**Task 1:** Added 80px padding to the bottom on the content panel (height of the nav), This bug only appears on my phone and I can't recreate it on my machine using dev tools or the browser plugin I have, so it needs live testing.

rogup:
@ms0ur1s I'll take task 4 and do it as part of my work on #64

lin-d-hop:
Guys! Impressive!
@rogup @ms0ur1s

lin-d-hop:
Can I confirm this has been deployed? Still seeing these issues on my phone :(

ms0ur1s:
> @ms0ur1s I'll take task 4 and do it as part of my work on #64

Thanks man ❤

rogup:
> Can I confirm this has been deployed? Still seeing these issues on my phone :(

Yes, deployed. I'm seeing the changes live - mabybe some browser caching on your end?

ms0ur1s:
### Andriod and IOS positional difference for reference

### Android
![Image](https://github.com/user-attachments/assets/d7670c5b-f4eb-4b96-906c-73487de93cbd)

### IOS
![Image](https://github.com/user-attachments/assets/553edb81-7ce7-411d-8baa-de476caf72ad)

ms0ur1s:
### [PR for the apply filters box positioning bug](https://github.com/DigitalCommons/mykomap-monolith/pull/88)

rogup:
@ColmDC I can't repro this. Maybe it was due to the confusion of the earlier deploy not having updated code. Is it still happening for you, and in what location on the map?

ColmDC:
I'm still seeing it on deploy-2024-11-19-41-g2e5b07b.

[...]

PR #73, opened 2024-11-20
Mobile UI bug fixes

ms0ur1s:
#### What? Why?

Related to issue #66

- Added padding to Content panel on mobile view to prevent content being hidden behind mobile nav
- Adjusted height of the mobile pop-up
- Added `Apply Filters` button to mobile render

- Check that when scrolling up on the mobile panels, that content at the bottom of the page is not hidden by the nav (requires a mobile check as the bug doesn't appear in dev tools)
- Check that the full pop-up is visible on a small mobile screen
- Check the `Apply Filters` button appears on mobile and that it is disabled until the search box is populated or a filtered has been selected.


rogup:
Nice work, I made a few more fixes, will merge now. Note how the translations work with the `t()` function

PR #88, opened 2024-11-22
[front-end] Add ApplyFilters component and integrate into filter Panel

ms0ur1s:
#### Mobile view
On the search screen the apply filters box was layered on to of the navigation

Related to issue #66

- Moved the apply filters box from the `SearchPanel` component to it's own `ApplyFilters` component.
- Added the `ApplyFilters` component to the `Panel` component
- Removed any related logic from the `SearchPanel` component and added it to the `Panel` component
- Tested across Andriod and IOS mobile devices

#### Checklist

- [ ] Added UTs

Using a mobile device of browserstack:
- Check that the apply filters box no longer appears on top of the navigation
- Check that the apply filters button still functions correctly

### DIFF
--- a/apps/front-end/src/components/map/mapLibre.ts
+++ b/apps/front-end/src/components/map/mapLibre.ts
@@ -172,16 +172,25 @@ export const createMap = (
         e: MapLayerMouseEvent,
         leafOffset: [number, number],
       ) => {
-        const coordinates = feature.geometry.coordinates.slice() as LngLatLike;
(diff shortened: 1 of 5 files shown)


rogup:
@ms0ur1s Nice work on this, it's looking great when I test it on my phone connected by USB :)

I added a slight fix for some funny behaviour due to the different uses of setText and setCurrentText

And also ended up accidentally adding other fixes to this branch, but it's fine since they're in separate commits. Will merge it all now

PR #90, opened 2024-11-22
Improve deploy doc

rogup:
Whilst deploying the application on prod-2, I updated the doc with the missing details I needed, and fixed bits that were incoreect

### DIFF
--- a/docs/deployment.md
+++ b/docs/deployment.md
@@ -153,6 +153,14 @@ The steps:
     
 
 ### Step 2: setup as the application user
+- The public SSH key of `$USER` has been added as a deploy key for
+  the [`cwm-test-data`](https://github.com/DigitalCommons/cwm-test-data/settings/keys)
+  private repository.
+- This key is used when connecting to github.com. We recommend configuring this in the
+  user's `~/.ssh/config` file, since you will may need fetch updates to the data on a regular
+  basis.
+
+The steps:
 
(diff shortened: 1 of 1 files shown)


docs/deployment.md
```diff
@@ -66,31 +66,41 @@ Examples, at the time of writing, of the typical case for these are:
...
-An example of the contents of a `$DEPLOY_ENV` file, but with secret
-values redacted:
+Here is an example of how the `$DEPLOY_ENV` file may be created as the
+application user, but with secret values redacted:
 
+    cat > $DEPLOY_ENV <<EOF
     export USERDIR=/home/$USER
-    export DEPLOY_DEST=$USERDIR/deploy
+    export GIT_WORKING=/home/$USER/gitworking/mykomap-monolith
+    export DEPLOY_DEST=/home/$USER/deploy
```
wu-lee, line 81: Note that in general Linux does not require home directories to be anywhere in particular. This is why I avoided hardwiring the `DEPLOY_DEST` directory to `/home/$USER/deploy` which makes the implicit assumption they'll be in `/home`, and put `$USERDIR/deploy` instead. (I am avoiding assuming a DCC server is being used as well)

wu-lee:
All seems ok to me, I think this can be merged

wu-lee:
rebased to the current head of `dev` before merging...

PR #93, opened 2024-11-23
[back-end] Add Glitchtip error tracking

rogup:
#### What? Why?

This will allow us to track uncaught exceptions and also 4xx and 5xx HTTP errors, viewable on the Glitchtip dashboard https://app.glitchtip.com/digital-commons-coop/issues?project=9203

(I included 4xx errors since we shouldn't really be hitting these if the FE is supplying the correct data)

It would be nice to notify Element devops channel but doesn't look like Glitchtip supports this yet. We could probably get it working eventually though, with some sort of workaround using webhooks or Matrix bots.

I've tested that this works on my local machine and the events are being picked up on the Glitchtip console

#### Deployment notes

Need to change `export GLITCHTIP_KEY=<key>` in the deploy.env to `export FE_GLITCHTIP_KEY=<key>;export BE_GLITCHTIP_KEY=<key>;` when deploying to the servers.

### DIFF
--- a/apps/back-end/src/errors.ts
+++ b/apps/back-end/src/errors.ts
@@ -1,0 +1,7 @@
+export class HttpError extends Error {
+  statusCode: number;
+  constructor(statusCode: number, message: string) {
+    super(message);
+    this.statusCode = statusCode;
+  }
+}
--- a/apps/back-end/src/services/Dataset.ts
+++ b/apps/back-end/src/services/Dataset.ts
@@ -1,16 +1,15 @@
 import fs from "node:fs";
 import path from "node:path";
 import _ from "lodash";
-import { TsRestResponseError } from "@ts-rest/core";
 import { z } from "zod";
 
 import {
-  contract,
   PropDefs,
   PropDefsFactory,
   schemas,
   TextSearch,
 } from "@mykomap/common";
+import { HttpError } from "../errors.js";
 
 // Infer type from Zod schema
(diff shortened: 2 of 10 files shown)


apps/back-end/src/errors.ts
wu-lee, line 1: Just curious - why a new error class rather than TsRestResponseError?

apps/back-end/src/services/Dataset.ts
```diff
@@ -90,12 +89,7 @@ export class Dataset {
 
   getItem = (itemIx: number) => {
     if (!fs.existsSync(path.join(this.folderPath, "items", `${itemIx}.json`))) {
-      throw new TsRestResponseError(contract.getDatasetItem, {
-        status: 404,
-        body: {
-          message: `can't retrieve data for dataset ${this.id} item @${itemIx}`,
-        },
-      });
+      throw new HttpError(404, `Can't retrieve data for dataset item`);
```
wu-lee, line 92: I suggest adding the ID back in, otherwise it makes it a bit harder to see if the ID isn't being interpreted correctly. (And at your suggestion, I'll add this in a bit.)

wu-lee:
Apart from the very minor points in the comments, this is all good - will merge it myself as you suggest,

apps/back-end/src/errors.ts
rogup: The TsRestResponseErrors skip middleware and serve them as a normal non-error response according to Fastify, meaning that the errors don't get sent to Glitchtip

Issue #103, opened 2025-01-24
[CWM] When zoom to co-op is slow pop-up doesn't pop


ColmDC:

King-Mob:
So far I can only replicate the FNCC part, which is caused by the location being inside a cluster marker even at the highest zoom level, because it shares a building with another location.

I've added code that automatically spiderfies and then opens the popup.

Recommend testing the first part (unreplicated) and getting a % likelihood that it happens

ColmDC:
I can no longer replicate this other symptom and it did not accur during @ms0ur1s 's  review, so let's just consider the last FNCC part.

ColmDC:
This is an annoying bug, as it seems very timing dependent. The FNCC example seems to work okay now, but I ma getting a search for Wisconsin Natural Heritage Cooperative failing consostently on my Firefox on Windows.
https://www.awesomescreenshot.com/video/38119240?key=5fbff03e6507badcfda9b38188015ba4

And on Chrome it finishes looking like this.

![Image](https://github.com/user-attachments/assets/8abeb1db-4311-4400-8a5b-581e16d2db17)

King-Mob:
As a result of the React bindings discussion, we decided to try out removing the recursive zoom function and just zooming to max depth for all markers when their item is clicked on in the directory

ColmDC:
I'd be keen to QA this soon, if someone can review the code.

King-Mob:
@ColmDC I just put it on dev

ColmDC:
Tested with https://dev.maps.coop/cwm/?datasetId=cwm-latest&lang=en

On Windows/Firefox. 

The behaviour seems a little better, but popups still occassionally fail to display. 

Are there explicit timing settings to play with here? Wondering if we should consider a different behaviour rather than just trying to fix this laboured pan across the planet?

King-Mob:
@ColmDC I'm up for moving away from laboured pan. I'll push a version with moveTo rather than flyTo in a few minutes

King-Mob:
@ColmDC have a go with that. I've stripped the logic right back

lin-d-hop:
In an ideal world every experiment will have its own issue so we can avoid this one issue exploding... but when is the world ideal?

King-Mob:
@lin-d-hop It doesn't really impact the React decision. I don't think that needs reopening. 

The question is what is the right behaviour for the map when a directory item is clicked on? 

[...]

PR #107, opened 2025-02-19
103 zoom popup open

King-Mob:
#### What? Why?

Related to issue #103

Locations in the same place were hidden by a cluster and wouldn't open as a popup when selected from the directory.

Needed new data points to test scenarios of zooming far and same location of markers.

#### Checklist

- [x] Added UTs

- Using `datasetId=dataset-A`, click on North Apples Coop and watch the map zoom in and open the cluster and then open a popup.

#### Deployment notes

No


King-Mob:
@rogup @wu-lee Got it down to just 1 test failing, which is the cli to dummy copy data test. Had a go at fixing it on the branch by changing the dummy data in the cli side. This didn't work. That section of the code is a bit inscrutable to me, could you take a look and tell me what is going wrong?

rogup:
Think @wu-lee is best placed to look at this, since he wrote that bit of code and tests

wu-lee:
Looks good to me code-wise - so far as I can tell the first couple of commits do something useful, then the others are just a mistake and more-or-less cancel each other out? (With minor changes to the data)

When I try the QA test, it works if I use separate back-end and front end invocations. 

(If I use `npm run start` from the front end, which runs the back-end too, it looks fine initially but if I open the directory panel and click "- any -", although it says "7 matching results", I don't see any organisations listed at all. I think this is a lingering bug with the latter, however.)

wu-lee:
@King-Mob - one afterthought, hopefully you'll see this.

We've been using a convention in our commit subjects to label the component being modified, so:
- back-end commits are prefixed with `[back-end] `
- front-end commits are prefixed with `[front-end] `
- common lib commits are prefixed with `[common] `

etc. 

The reason: as these are essentially separate applications or libraries wrapped up as a monolith repository, we wanted to make it clear which one is being altered, and remind ourselves to avoid altering several at once.

As well as helping the reviewer, this is in case one or more of these components needs to be split out as its own repository later, in which case we want the descriptions make sense standing alone within that component if the prefix is removed.

Issue #55, opened 2024-11-18
[CWM] Add pagination to ResultsPanel


rogup:
## Description

We currently fetch all names matching a search to populate the ResultsPanel. This will probably become a performance issue if the dataset gets too big.

The search API supports pagination so we just need to implement this in the UI, maybe using this MUI component:
https://mui.com/material-ui/react-pagination/#controlled-pagination

## Acceptance Criteria
- [ ]

rogup:
@ColmDC No, for now, I've just truncated the list to 200 so loading isn't too slow, and the user can't see past the first 200 items unless they refine their search https://github.com/DigitalCommons/mykomap-monolith/commit/f4b295f037f2b1d6434e9776d655e3331ed652ca

ColmDC:
Can you estimate how big a task this would be?

rogup:
I reckon 1-1.5 days: 0.5-1 day to design and create the pagination UI component, then half a day to link it up to the API and write Redux UTs

ColmDC:
Is this one you had started @King-Mob, but put back in Dev Ready for a reason? Is so, what was the reason?

King-Mob:
@ColmDC reason is the final piece of rohit's review, the width of the page numbers

ColmDC:
> reason is the final piece of rohit's review, the width of the page numbers

Sorry, I'm confused. Please create link here to review you reference.

King-Mob:
Here's the conversation about page numbers width: https://github.com/DigitalCommons/mykomap-monolith/pull/109#discussion_r1979310825

ColmDC:
Where can I QA this one?

ColmDC:
> Where can I QA this one?

Can you check that this feature was pushed to somewhere I can QA it?

King-Mob:
@ColmDC Up on dev.maps.coop

(wasn't there previously because was merged into main rather than dev)

ColmDC:
Nice.

PR #109, opened 2025-03-03
add paginated search results

King-Mob:
#### What? Why?

Related to issue #55 

Following setSelectedTab for redux example, and using the values that were already in the code marked TODO, I added the pagination component from mui.

#### Checklist

- [ ] Added UTs
- [X] Checked that all automated tests pass

- Open directory
- Scroll to bottom
- Click on the pages to change which page you view

#### Deployment notes

No

### DIFF
--- a/apps/front-end/src/components/panel/panelSlice.ts
+++ b/apps/front-end/src/components/panel/panelSlice.ts
@@ -4,7 +4,7 @@ import { getDatasetId } from "../../utils/window-utils";
 import { searchDataset } from "../../services";
 import { SearchSliceState } from "./searchPanel/searchSlice";
 
-export const RESULTS_PER_PAGE = 200;
+export const RESULTS_PER_PAGE = 2;
 
 interface PanelSliceState {
   selectedTab: number;
@@ -99,10 +99,7 @@ export const panelSlice = createAppSlice({
           console.error("Error fetching search results", action.payload);
         },
       },
-    ),
-    setResultsPage: create.reducer((state, action: PayloadAction<number>) => {
-      state.resultsPage = action.payload;
-    }),
+    )
   }),
   selectors: {
     selectSelectedTab: (panel) => panel.selectedTab,
@@ -123,7 +120,6 @@ export const {
   openResultsPanel,
   closeResultsPanel,
   populateSearchResults,
-  setResultsPage
 } = panelSlice.actions;
 
 export const {
--- a/apps/front-end/src/components/panel/panelSlice.ts
+++ b/apps/front-end/src/components/panel/panelSlice.ts
@@ -4,7 +4,7 @@ import { getDatasetId } from "../../utils/window-utils";
 import { searchDataset } from "../../services";
 import { SearchSliceState } from "./searchPanel/searchSlice";
 
-const RESULTS_PER_PAGE = 10;
+export const RESULTS_PER_PAGE = 200;
 
 interface PanelSliceState {
   selectedTab: number;
@@ -68,8 +68,8 @@ export const panelSlice = createAppSlice({
           query: {
             ...search.searchQuery,
             returnProps: ["name"],
-            page: 0, // TODO: implement pagination. For now, just fetch first 200 results
-            pageSize: 200, // RESULTS_PER_PAGE,
+            page: page,
+            pageSize: RESULTS_PER_PAGE
           },
         });
         if (response.status === 200) {
@@ -100,6 +100,9 @@ export const panelSlice = createAppSlice({
         },
       },
     ),
+    setResultsPage: create.reducer((state, action: PayloadAction<number>) => {
+      state.resultsPage = action.payload;
+    }),
   }),
   selectors: {
     selectSelectedTab: (panel) => panel.selectedTab,
@@ -120,6 +123,7 @@ export const {
   openResultsPanel,
   closeResultsPanel,
   populateSearchResults,
+  setResultsPage
 } = panelSlice.actions;
 
 export const {
--- a/apps/front-end/src/components/panel/panelSlice.ts
+++ b/apps/front-end/src/components/panel/panelSlice.ts
@@ -4,7 +4,7 @@ import { getDatasetId } from "../../utils/window-utils";
 import { searchDataset } from "../../services";
 import { SearchSliceState } from "./searchPanel/searchSlice";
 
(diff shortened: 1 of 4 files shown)


King-Mob:
@rogup noticed on the issue that there was the intention of adding redux tests. I've skipped this on the justification that what I've added on the redux side is a "humble method".

apps/front-end/src/components/panel/resultsPanel/results/Results.tsx
```diff
@@ -79,6 +86,16 @@ const Results = () => {
           />
         ))}
       </List>
+      {totalPages > 1 && (
+        <Pagination
```
rogup, line 90: The styling of this component means it's quite hard to see (and probably doesn't pass accessibility contrast guidelines)
Screenshot 2025-03-04 at 12 10 14

rogup: Also, I think the width needs to be adjusted somehow so it doesn't wrap onto multiple lines when there are larger page numbers

apps/front-end/src/components/panel/resultsPanel/results/Results.tsx
```diff
@@ -79,6 +86,16 @@ const Results = () => {
...
+          count={Math.ceil(resultCount / RESULTS_PER_PAGE)}
```
rogup, line 91: can reuse totalPages variable

apps/front-end/src/components/panel/resultsPanel/results/Results.tsx
rogup: Also, I think having it fixed at the bottom of the results panel, outside of the scrollable region would be nicer. So it's always visible, doesn't jump around based on the wrapping/number of the actual results, and so you don't need to scroll for a long time in order to reach the pagination control

And similarly, the "X matching results" at the top of the results list may need to be separated. Or it could be kept in the same place and the text could be changed to something like "Results 1-100 of X", in the case of pagination

apps/front-end/src/components/panel/resultsPanel/results/Results.tsx
```diff
@@ -79,6 +86,16 @@ const Results = () => {
...
+          page={resultsPage + 1}
+          onChange={(event, value) => {
+            dispatch(setResultsPage(value - 1));
+            dispatch(performSearch());
```
rogup, line 95: I think this can just call `populateSearchResults` directly, rather than doing a whole search (which re-renders the markers too)

apps/front-end/src/components/panel/panelSlice.ts
```diff
@@ -100,6 +100,9 @@ export const panelSlice = createAppSlice({
         },
       },
     ),
+    setResultsPage: create.reducer((state, action: PayloadAction<number>) => {
```
rogup, line 103: The above `populateSearchResults` method sets the results page when it is fulfilled, so I don't think we need this as a separate method, we can just call `populateSearchResults` from the React component

apps/front-end/src/components/panel/searchPanel/searchSlice.ts
```diff
@@ -182,13 +182,14 @@ export const performSearch = (): AppThunk => {
...
-    const { search } = getState();
+    const { search, panel } = getState();
+
     const activeFilters = search.filterableVocabProps.filter(
       (prop) => prop.value !== PROP_VALUE_ANY,
     );
     if (activeFilters.length === 0 && search.text === "") {
       dispatch(updateVisibleIndexes({ searchQuery: {}, visibleIndexes: [] }));
-      dispatch(populateSearchResults(0));
+      dispatch(populateSearchResults(panel.resultsPage));
```
rogup, line 192: I think this should still be `populateSearchResults(0)` since when we perform a new search with new filters, we want pagination to reset.
See my other comment about not calling this function if we only want to populate search results for the same search query

apps/front-end/src/components/panel/searchPanel/searchSlice.ts
```diff
@@ -212,7 +213,7 @@ export const performSearch = (): AppThunk => {
           ),
         }),
       );
-      dispatch(populateSearchResults(0));
+      dispatch(populateSearchResults(panel.resultsPage));
```
rogup, line 216: ditto

apps/front-end/src/components/panel/panelSlice.ts
```diff
@@ -4,7 +4,7 @@ import { getDatasetId } from "../../utils/window-utils";
 import { searchDataset } from "../../services";
 import { SearchSliceState } from "./searchPanel/searchSlice";
 
-const RESULTS_PER_PAGE = 10;
+export const RESULTS_PER_PAGE = 200;
```
rogup, line 7: A bit less scrolling per page might be nice, and would also reduce load on the server (since someone is unlikely to look through 200 results before refining a search). Maybe 50 or 100 would be good?

rogup (commented):
@King-Mob Generally looking good - I've left a few comments based on things I noticed after doing a quick test of the code in my dev environment, some more pedantic than others but I think worth fixing

apps/front-end/src/components/panel/resultsPanel/results/Results.tsx
King-Mob: @rogup for the width, can you tell me where to get a dataset large enough to trigger that? at the moment i'm just using dataset A, which has 3 items in it

King-Mob: @rogup width is the only thing left to do (I think. Hard to judge the fixed stuff without the larger dataset)

rogup: @King-Mob did you figure out how to use the Delhi dataset in testing?

King-Mob: Yes, been focusing on the popup zoom bug though

King-Mob: @rogup in the delivery call last week we talked a bit about the width issue and the suggestion was to let it get through to QA because it's not looking like a quick fix. Am I alright to resolve this?

rogup: @King-Mob Sure

Issue #67, opened 2024-11-20
[CWM]  Domains should list in 2 columns when there are more than 10


lin-d-hop:
## Description

As per the [designs](https://xd.adobe.com/view/3ef8c8d0-84bb-480e-bdfe-113d11f1583b-6bab/screen/2cccf667-44b9-4e12-ac7e-3c1c9d17eb60/), when there are more than 10 domains listed, we need these to list over 2 columns.

![Image](https://github.com/user-attachments/assets/8c2848f6-bc9e-44ae-98db-286fda2cc3fa)

## Acceptance Criteria

- [x] Domains displayed in two columns when there are more than 10
- [x] When displaying in 2 columns there domains should split evenly over the two columns - 5 and 5 for example
- [x] When the domain char length is greater than the possible width that can be displayed on one line, cut the domain short and add an ellipses (...). The link should still work in this case.

ColmDC:
Keen to QA this today. :-)

King-Mob:
@ColmDC up on dev.maps.coop

King-Mob:
Change up on dev. For screens over 400px I've increased the width of popup to 380px. For screens below 400px I've listed the domains in one column.

PR #110, opened 2025-03-03
67 use two columns over ten domains

King-Mob:
#### What? Why?

Related to issue #67

#### Checklist

- [X] Added UTs
- [X] Checked that all automated tests pass

- Open an item with more than 9 domains
- see the 2 columns

#### Deployment notes

No

### DIFF
--- a/apps/front-end/src/components/popup/leftPane/LeftPane.tsx
+++ b/apps/front-end/src/components/popup/leftPane/LeftPane.tsx
@@ -103,11 +103,13 @@ const LeftPane = ({
   const cleanDescription = decode(
     description?.replace(/<[^<>]+(>)/g, ""), // remove HTML tags
   )?.trim();
+
   const MAX_DOMAIN_LENGTH = 25;
+  const DOMAINS_SINGLE_COlUMN = 9;
 
   //split into 2 columns if more than 10 domains
   const dcDomains =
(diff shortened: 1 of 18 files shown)


wu-lee:
When I try to test this locally with dataset-A, all pop-ups seem to explode the app, with this error:

> TypeError: values.map is not a function

Reported in Glitchtip here:

https://app.glitchtip.com/digital-commons-coop/issues/2674393

apps/front-end/src/components/popup/leftPane/LeftPane.tsx
```diff
@@ -103,11 +103,13 @@ const LeftPane = ({
   const cleanDescription = decode(
     description?.replace(/<[^<>]+(>)/g, ""), // remove HTML tags
   )?.trim();
+
   const MAX_DOMAIN_LENGTH = 25;
+  const DOMAINS_SINGLE_COlUMN = 9;
```
wu-lee, line 108: Did you mean to spell column in mixed case?

wu-lee:
Looks good.

For future work, the magic constants setting the thresholds might need to be tweaked on  different projects, so I suspect they belong in some configurable setting.  But I think we don't want to do that now. Also I still haven't worked out how we do things like that in this version of MM.

PR #111, opened 2025-03-06
Fix `npm run start` in back-end app

wu-lee:
#### What? Why?

There is a run script in the MM front-end application's `package.json` called: `sta
rt`.

    npm run start

This should launch a server on http://localhost:3000 which has both front and back end integrated on the same URL, as it is in a production deployment. The front-end is accessible via the above URL, and the back-end via http://localhost:3000/api/.

Obviously it should work functionally identically to the live deployment. However it has not been: when viewing the directory, the listings for markers were not being shown, despite the total number of items being shown correctly.

This is a bug.

Note however, there is an alternative development process in which two processes are run:

    npm run dev # in the back end, listens on http://localhost:3000

And

    npm run dev # in the front end, listens on http://localhost:5173

This sets `VITE_API_URL=https://localhost:3000` in the environment of the front end to make it use the different back-end URL, and it requires the back-end to have CORS enabled to permit this to work.

This method is not exhibiting the bug. Nor are the production/development deploys on the server.

#### To replicate

To replicate:
- Check out the MM repository as normal, and `npm install` at the top level
- In `src/front-end` run `npm run start`
- Visit http://localhost:3000/?datasetId=dataset-A
- Check that there are some markers on the map - shoud be 4 of them.
- Open the side panel
- Click on "- any -"
- You should see the words "7 Matching results", but no items in the search results.

What should happen:
- As above, but all seven results should be shown

Check that the code is readable and comprehensible.

#### Checklist

I'm not sure of a way to make a automated test for this. Nevertheless:

- [ ] Check that the alternative method still works (see procedure above)

Follow the replication procedure above. It should now give the expected outcome.

### DIFF
--- a/apps/back-end/src/index.ts
+++ b/apps/back-end/src/index.ts
@@ -1,6 +1,6 @@
 import fastifySentryPlugin from "@immobiliarelabs/fastify-sentry";
 import closeWithGrace from "close-with-grace";
-import Fastify from "fastify";
+import { FastifyListenOptions, fastify } from "fastify";
 import cors from "@fastify/cors";
 import qs from "qs";
 import apiPlugin from "./pluginApi.js";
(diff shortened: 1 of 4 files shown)


wu-lee:
An analysis of what goes wrong.

The `start` runscript in `apps/front-end/package.json` which launches the back end and the front end as a unified service uses `apps/front-end/server.ts` as an entry point. That in turn uses `@fastify/vite`, to launch the back-end API as a fastify plugin, defined in `apps/back-end/src/pluginApi.ts`. 

The `dev` and `start` runscripts in `apps/back-end/package.json` launch the back end alone use `apps/back-end/start.[tj]s` as an entry point. This calls a function `start` in `apps/back-end/src/index.ts`, and in turn this loads the back-end API as a fastify plugin, defined in `apps/back-end/src/pluginApi.ts`. 

Evidently the code path for these two entry points are different.

This manifests as the back-end's API endpoint for searching ignoring the `returnProps` parameter in the former case, which means only the item indexes are returned, with any name.

So this call:

    http://localhost:3000/api/dataset/dataset-A/search?returnProps%5B0%5D=name&page=0&pageSize=200

Returns this:

```
[
  {
    "index": "@0"
  },
  {
    "index": "@1"
  },
  {
    "index": "@2"
  },
  {
    "index": "@3"
  },
  {
    "index": "@4"
  },
  {
    "index": "@5"
  },
  {
    "index": "@6"
  }
]
```

Instead of this:

```
[
  {
    "index": "@0",
    "name": "Apples Co-op"
  },
  {
    "index": "@1",
    "name": "Pears United"
  },
  {
    "index": "@2",
    "name": "Invisible Collab"
  },
  {
    "index": "@3",
    "name": "Kangaroo Koop"
  },
  {
    "index": "@4",
    "name": "Koala Koop"
  },
  {
    "index": "@5",
    "name": "North Apples Co-op"
  },
  {
    "index": "@6",
    "name": "North Apples Co-op 2"
  }
]
```

The reason for this appears to be that the latter mechanism has the `qs` query string parsing extension loaded in the fastify server. This understands the style of array parameter used in the API call. But the former mechanism does not load that.

So the solution is to unify the two to use the same code-path, ensuring the server is provisioned the same in both cases.

rogup (approved):
Looks good

Issue #116, opened 2025-03-13
[CWM] [BUG] Spidered Markers Do Not Toggle


ms0ur1s:
### Description
Single locations with multiple markers appear as clusters and when clicking on such clusters the markers spider out. However, if a pop-up is opened from a spidered marker, the pop-up does not close when the marker is clicked again. &mdash; usual behaviour for a marker within the CWM. This missing functionality can things especially awkward when dealing with [large clusterings of makers.
](https://www.awesomescreenshot.com/video/36872096?key=2d36f89fc3f48a3361507966b77ab418)

**Steps to reproduce**
  - Load map
  - Open the directory tab
  - Click on Argentina
  - Click on SanCor Cooperativas Unidas Limitada
  - Click on the cluster
  - The marker [does not toggle](https://www.awesomescreenshot.com/video/36834763?key=d066c7f89b8b3706503accd8e44be6ce)

### Expected Behaviour
- Clicking a spidered marker should toggle the pop-up.

***

**Notes**
- CWM UI Review issue can be found [here](https://github.com/DigitalCommons/mykomap-monolith/issues/104)

ColmDC:
Can you check how the clustering behaves after this change with the new Barcelona data set?

datasetId=cwm-20250324

It has a set of 10 clustered co-ops.

ColmDC:
From a Github usage pov, should this ticket be closed while still in QA?

ColmDC:
> I agree with you. Currently it closes automatically when I merge a linked pull request.

Where would that be set? @lin-d-hop ?

ColmDC:
Looks like the geocoding has changed, so this example no longer results in clustered co-ops. So no longer a good test.

ms0ur1s:
@ColmDC, perhaps this change hadn't propagated to dev when you tested it. I've tested desktop on chrome and firefox and it works for me (see below). Will just give it a test on couple of other devices to make sure.

https://github.com/user-attachments/assets/82c06980-988e-4058-bf3a-4f908e62d815

ColmDC:
>Looks like the geocoding has changed, so this example no longer results in clustered co-ops. So no longer a good test.

What data set are you testing this on? My point was that with the new data, those 3 co-ops are no longer co-located. Sancor Cooperativa de Seguros Ltda is now positioned away from the square on its own, so that example is no longer useful for illustrating the issue with the spidering. But the bug has been fixed, so all good.

ms0ur1s:
[...]

PR #121, opened 2025-03-19
[front-end] mapLibre.ts - Enhance popup handling and cursor styling for spiderfied markers

ms0ur1s:
#### What? Why?

Related to issue #116

Markers connected to spiderfied clusters were not toggling their respective pop-ups. If a cluster was clicked the pop-up would open, if clicked again it would close and then immediately re-open. Expected behaviour being if a marker is clicked it should toggle its related pop-up either open or closed.

Related to issue #116

Changes to the `createMap` function in the `apps/front-end/src/components/map/mapLibre.ts`, lines 184 -192 and 219

- Added logic to toggle the popup when it is already open for the same item, including logging and cleanup.
- Added an event listener to change the cursor to a pointer when hovering over a spiderfied cluster.

- Navigate to location that contains multiple co-ops
- Click on the cluster
- Hover over a marker and ensure you see a pointer cursor
- Click on the marker to trigger the pop-up
- Click on the marker to ensure the pop-up closes and remains closed until toggled again

### DIFF
--- a/apps/front-end/src/components/map/mapLibre.ts
+++ b/apps/front-end/src/components/map/mapLibre.ts
@@ -180,6 +180,17 @@ export const createMap = (
       ) => {
         const coordinates = feature.geometry.coordinates.slice();
         const itemIx = feature.properties?.ix;
+
+        if (popup?.isOpen() && popupIx === itemIx) {
+          console.log(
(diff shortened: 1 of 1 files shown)


rogup (commented):
Looks good and ready to merge

Issue #95, opened 2024-12-02
[CWM] Source maps missing from Glitchtip


rogup:
**Describe the bug**
On the front-end error reports on Glitchtip (e.g. https://app.glitchtip.com/digital-commons-coop/issues/1775542), the stacktraces are not readable since the code is minified and we haven't given Glitchtip the sourcemaps.

Adding sourcemaps is apparently supported by Glitchtip using the sentry-cli, which we can maybe add to our deploy script
https://gitlab.com/glitchtip/glitchtip/-/issues/13
https://glitchtip.com/blog/2022-04-25-glitchtip-1-12
https://gitlab.com/glitchtip/glitchtip-backend/-/issues/322 <-- a recent issue which might be helpful

Fixing this will allow us to analyse errors and identify bugs much more easily.

**Expected behaviour**
We can see the un-minified stacktrace of front-end errors on Glitchtip

ColmDC:
Are you working on this @wu-lee ? If so, move into In Progress? Is it general MM tech debt or particular to CWM. ?

ColmDC:
Should it be labvelled Tech Debt?

wu-lee:
PR added: #129 . Review requested.

Deploying it has been tested on dev-2. There's probably some conversations to be had about our release tagging process which come out of this, because the solution requires consistent release tags to be applied to code and their source-maps.

PR #129, opened 2025-03-25
95 glitchtip sourcemaps

wu-lee:
#### What? Why?

Related to issue #95

As explained in the issue: front-end stack traces in Glitchtip's logs don't refer to any readable code because it is minimised, but they could if the sourcemap files were submitted to Glitchtip correctly on deployment. 

#### Additional notes

Also, it turns out: 
- The source map files also seem to need to be deployed on the front-end server, or legible stack traces are not shown correctly (this does happen normally, but not when running locally in development).
- The front-end Sentry API initialisation call needs to specify a _revision label_ in its paramters, which can be matched to the same revision label (and a filename) on a sourcemap submitted to Glitchtip.
- Traces from the back-end code currently do not need this sourcemap upload: the stack traces seem to be legible already.

Nevertheless, in order to be able to supply a revision label like this, it requires a consistent labelling of builds, including the version number.  Which needs to come from somewhere - the conventional way these days seems to be to use release tags in version control.

For identification of the exact build to inspect when an error is reported, it requires a build number as well.  The front end module at least should have this labelling - but for consistency also the back end, and perhaps all the modules need it.

Getting this right consistently across the modules in the project is the main component of this PR.

Additionally, if we are tagging and labelling, the `package.json` files also contain a version number which should be updated consistently as well.  

`git-describe` is the usual command to get the last tag and other metrics from the git commit history, which can be used for the build information. This needs to be run, parsed, and the labels constructed from that.

Our build framework is Vite. This allows build-time constants to be defined, which can then be inserted into the compiled source as literals and emitted at runtime, with some caveats (e.g. this constant must be stringifiable).  We were already using this mechanism for the API endpoint which reports the back-end version. It needs to be employed in the front-end too. For consistency it's added in all the modules, using the a common mechanism, avoiding duplication.

Finally, the second point in the list above means: the deployment process needs to be adjusted to ensure that sourcemaps for the front end (at least) are submitted to Glitchtip, with the correct release label.

Getting  consistently working solution took a number of iterations to experiment with different approaches and configurations and find one which worked. These have been tidied up via a number of rebases so that the result is more intelligible for the reviewer. It may not be entirely perfect, so check for evidence that anything has been missed.

The builds and tests have been run and checked for all the significant commits where this would be expected to work, post rebase.

All the modules are now labelled in a way which should allow the build information to be inserted into the code at build time, and thereby reported via API calls etc. at runtime.  

There are some consequences from the changes in this PR, in particular because the `package.json` files' versions are now being updated from the commit history tags during the build.

- It is necessary to tag the commits using a tag with the form `v<A>.<B>.<C>`, where A B and C are positive integers. (Three is used in this example, but more can be used; there can be at minimum one number. Tags not matching this format should be ignored, e.g. `v100-dalmations`, `v.1`, etc.)
- This numeric sequence is written into the appropriate `package.json` attribute `version` on building. This then needs to be committed by the developer.
- The front-end module's `package.json` also gets an attribute written, `config.sentry.release`, that indicates the release label in a format the Sentry API permits.
- This release label is used by the `npm run upload-sourcemaps` runscript to tag the sourcemaps appropriately.

We might want to discuss the consequences and the processes that they imply.

In particular, when to do the tagging, committing and uploading needed for this to work is open to question. 

Release tagging and committing: the main time this is needed is on the main branch to indicate releases. Version tags seem to be most appropriate there. A rebuild will then be needed to update the versions in the `package.json` files. These should be committed. It seems difficult to do that on the tagged commit itself. Perhaps versions should be updated after, to avoid complication of amending the commit and re-tagging?

Uploading of source-maps: when to do this is more clear... it should be when the application is deployed. The deploy script has been altered to perform this step, after a rebuild.

Note that any alterations to the code made by the build will result in a `-dirty` tag appended to the build information, this deliberate as we probably want to avoid dirty builds. But it also means we need to take care that the build has been performed and the labels in the `package.json` files have been committed, before deployment.

Instructions for following the processes described above should be of course be documented. But I suggest that it can only be following discussions and agreement, and so it should be the subject for a later PR.

For now I've added some documentation explaining:
- the build sequence, reasons why it what it is, in `docs/monorepo.md`
- the extra environment variables needed for sourcemap deployment, in the `deploy.sh` script itself.

Unit testing the build process seems infeasible, so I think those points do not apply here. But perhaps how to test the build and deploy process can also be discussed.

#### Checklist

- [ ] Added UTs

- Check out the code
- Run `npm install`
- Run `npm run build` - should succeed
- Run `npm run test` - should succeed, and all modules tests should have run

#### Deployment notes

- Deployment process should be comprehensible from the documentation and comments
- Details of the sensitive variables needed for this and other deployment parameters should be retained somewhere shared but secure (not just on the server)
- Deployment should have been tested successfully on the development server, such that deployment on the production server is likely to be successful.

The second point may need some discussion, AFAIK it's not currently being met.  Possibly also the first, although as before I suggest alterations to documents should go into a new PR.

I've tested the last point personally, not sure if it needs to be repeated, but obviously that would be the most thorough sanity check.


rogup (commented):
Hi @wu-lee , I'm not going to do a thorough review on this, because it looks like there a lot of changes to get my head around, and I'm not sure how valuable it is for me to understand every single part of it.

I think I get the gist of what is happening. I was kind of imagining something really simple, like setting the source maps version to just the Git commit when deploying and then uploading them, and this being separate from the release versioning. But it's cool that you've found a way to make everything consistent!

Just one comment really, which is related to something you said:

> But it also means we need to take care that the build has been performed and the labels in the package.json files have been committed, before deployment.

I think we should include this in the docs in an appropriate place, even if we haven't agreed on a final way to do versioning, since it's an important piece of info. It might be a while before the next issue to decide this is scheduled, and I feel like its priority is lower than lots of our other tech debt.

I'd be in favour of you just outlining the process for now, and a suggestion for how versioning happens, and to go with this until we think it needs changing - it doesn't need to be perfect.

lin-d-hop:
Thanks team.
So is the next step updating the appropriate docs and closing this one?
Feels like its already gotten a bit unwieldy and perhaps is a lower priority than a lot of our other work so I'd be in favour of wrapping it up :)

rogup:
@lin-d-hop Yes I think that's all that's left

I'm wondering if we should have some kind of defined DCC check-in process for when a piece of works starts ballooning?

Like, maybe pausing and listing the different options for how to tackle the issue with rough time estimates for each approach so someone can review and weigh up the pros and cons. Kind of similar to what ended up happening on the unregistered layer issue but even earlier in the process if possible.

wu-lee:
I've added a proposal in docs/versioning-and-releases.md.  I'm going to merge this so that we can mark it done, and assume any discussion arising from that document can be addressed in an upcoming chunk of work.

lin-d-hop:
TODO Document supporting WSL Mac and Linux.
Then clooooooose

King-Mob:
@lin-d-hop yes, can do

King-Mob:
Having read this, I think the task is to judge the suitability of the proposal for versioning outlined here: https://github.com/DigitalCommons/mykomap-monolith/blob/dev/docs/versioning-and-releases.md

The proposal is to add a step before a production release, where the deployer runs "npm run release v<version-number>", and that's the versioning system as it relates to deploys.

I don't see a reason not to use the system as it has been outlined. It satisfies the criteria in a straightforward way.

I judge this to be suitable.

Issue #119, opened 2025-03-13
[CWM] Empty Directory Items Should Indicate Unpopulated Entries


ms0ur1s:
### Description
Until reaching Argentina, every directory entry returns a 'No matching results' message. This can make users assume the directory is broken or unpopulated. A more user-friendly approach is needed.

**Steps to Reproduce**
- Select the directory pane.
- Click on Afghanistan.
- Observe no matching results.
- Continue selecting countries until reaching Argentina.

### Suggested Solutions

1. Grey out empty directory entries and make them unclickable.
2. Display an item count next to each directory entry (e.g. Afghanistan (0)).

***

**Notes**
- This should align with existing issue [#98: Evaluate greying out titles of categories for which no entities match](https://github.com/DigitalCommons/mykomap-monolith/issues/98).
- CWM UI Review issue can be found [here](https://github.com/DigitalCommons/mykomap-monolith/issues/104)
- Implement solution 1. If when doing so it seems trivial to also do 2 then do, so, otherwise create another Issue to capture that feature.

ColmDC:
I'm keen to QA this today. :-)

PR #135, opened 2025-03-27
119 grey out empty directory entries

King-Mob:
#### What? Why?

Related to issue #119

Added 

Back-end code, specifically `back-end/src/services/Dataset.ts` line 119

This seems fragile to me. I've gotten the country code from the item json because it's in the index 1 position on the array. Is this the best way currently available to do this?

#### Checklist

- [ ] Added UTs

- Open directory
- See the numbers and greyed out where it's 0.

#### Deployment notes

rerun common building

### DIFF
--- a/apps/back-end/src/services/Dataset.ts
+++ b/apps/back-end/src/services/Dataset.ts
@@ -111,10 +111,19 @@ export class Dataset {
     fs.createReadStream(path.join(this.folderPath, "locations.json"), "utf8");
 
   getTotals = () => {
-    const totals = {};
+    const totals: any = {
+      any: 0
+    };
 
-    this.searchablePropValues.forEach((itemValues, itemIx) => {
-      console.log(itemValues, itemIx)
+    this.searchablePropValues.forEach((itemValues) => {
+      if (totals[itemValues[1] as string]) {
+        totals[itemValues[1] as string] = totals[itemValues[1] as string] + 1;
+      }
+      else {
+        totals[itemValues[1] as string] = 1;
+      }
+
+      totals.any = totals.any + 1;
     })
 
     return totals;
--- a/apps/back-end/src/services/Dataset.ts
+++ b/apps/back-end/src/services/Dataset.ts
@@ -4,6 +4,7 @@ import _ from "lodash";
 import { z } from "zod";
 
 import {
+  Dictionary,
   PropDefs,
   PropDefsFactory,
   schemas,
@@ -111,19 +112,15 @@ export class Dataset {
     fs.createReadStream(path.join(this.folderPath, "locations.json"), "utf8");
 
   getTotals = () => {
-    const totals: any = {
+    const totals: Dictionary<number> = {
       any: 0
     };
 
     this.searchablePropValues.forEach((itemValues) => {
-      if (totals[itemValues[1] as string]) {
-        totals[itemValues[1] as string] = totals[itemValues[1] as string] + 1;
-      }
-      else {
-        totals[itemValues[1] as string] = 1;
-      }
-
-      totals.any = totals.any + 1;
+      const label = itemValues[1];
+      if (typeof label !== "string") return;
+      totals[label] = 1 + (totals[label] ?? 0);
+      totals.any = 1 + (totals.any ?? 0);
     })
 
     return totals;
--- a/apps/back-end/src/services/Dataset.ts
+++ b/apps/back-end/src/services/Dataset.ts
@@ -81,8 +81,8 @@ export class Dataset {
     } else {
       throw new Error(
         `searchable.json for dataset ${this.id} has a bad format ` +
-          `(hasSearchStringField: ${hasSearchStringField}, uniqueItemProps: ${uniqueItemProps}, ` +
-          `sameItemPropsAsConfig: ${sameItemPropsAsConfig}, expectedValuesLengths: ${expectedValuesLengths})`,
+        `(hasSearchStringField: ${hasSearchStringField}, uniqueItemProps: ${uniqueItemProps}, ` +
+        `sameItemPropsAsConfig: ${sameItemPropsAsConfig}, expectedValuesLengths: ${expectedValuesLengths})`,
       );
     }
   }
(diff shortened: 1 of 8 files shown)


wu-lee:
The two commits seem to reverse each other in various places - I infer some of the first commit is experimental no-ops which just print totals on the console, and the second removes those and inserts the implementation?

Perhaps they deserve to be squashed into one commit? If I check the result of that, it looks simpler.

wu-lee:
(Above rebase is just to move the branch onto the end of `dev` BTW, applies cleanly with no other changes)

apps/back-end/src/services/Dataset.ts
```diff
@@ -111,10 +111,19 @@ export class Dataset {
...
-    const totals = {};
-
-    this.searchablePropValues.forEach((itemValues, itemIx) => {
-      console.log(itemValues, itemIx)
+    const totals: any = {
+      any: 0
+    };
+
+    this.searchablePropValues.forEach((itemValues) => {
+      if (totals[itemValues[1] as string]) {
```
wu-lee, line 119: This doesn't seem to deal with the possibility that `itemValues[1]` might be an `string[]`? Or that the array might have length 0?

The types are pesky. But this seems to keep the checker happy:
```
    this.searchablePropValues.forEach((itemValues) => {
      const label = itemValues[1];
      if (typeof label !== "string") return;
      totals[label] = 1 + (totals[label] ?? 0);
      totals.any = 1 + (totals.any ?? 0);
    });
```

apps/front-end/src/components/panel/directoryPanel/DirectoryPanel.tsx
```diff
@@ -25,12 +24,12 @@ const StyledDirectoryPanel = styled(Box)(() => ({
...
 
 const fetchResults = async () => {
-  const response = await searchDataset({
-    params: { datasetId: "delhi" },
-    query: {
-      returnProps: ["country_id"],
-    },
-  });
+  const datasetId = getDatasetId();
+  if (datasetId === null) {
```
wu-lee, line 28: I don't entirely trust `datasetId` won't be `undefined` here, because TS seems to have a bit of a blind spot for it...

apps/front-end/src/components/panel/directoryPanel/DirectoryPanel.tsx
```diff
@@ -25,12 +24,12 @@ const StyledDirectoryPanel = styled(Box)(() => ({
 }));
 
 const fetchResults = async () => {
```
wu-lee, line 26: Perhaps this should have a return type?

Or better: a switch like this allows the type of the body to be inferred correctly, and you can also deal with the errors, perhaps by throwing, or returning `{failed: true}` in a type-safe way.

```
const fetchResults = async () => {
  const datasetId = getDatasetId();
  if (datasetId === null) {
    return { failed: true };
  }

  const response = await getTotals({ params: { datasetId } });

  switch(response.status) {
    case 200: {
      const body = response.body; // Now has the correct type of the response in this case
      return body;
    }
      
    case 400:
      throw new Error("some error here "+response.body.message);
      
    case 404:
      throw new Error("some error here "+response.body.message);
  }
};
```
TS should spot any missing switch clauses... hopefully!

apps/front-end/src/components/panel/directoryPanel/DirectoryPanel.tsx
```diff
@@ -40,17 +39,12 @@ const DirectoryPanel = () => {
 
   const directoryOptions = useAppSelector(selectDirectoryOptions);
   const activeValue = directoryOptions.value;
-
-  const [resultsTotals, setResultsTotals] = useState<number[]>([]);
+  const [resultsTotals, setResultsTotals] = useState<any>();
```
wu-lee, line 42: If you use a switch statement above this doesn't need to be `<any>`, it can be something more appropriate like `Record<string, number>` (depending on what you return from `fetchResults`)

apps/front-end/src/components/panel/directoryPanel/DirectoryPanel.tsx
```diff
@@ -40,17 +39,12 @@ const DirectoryPanel = () => {
...
 
   useEffect(() => {
-    //this is where i will manipulate my list of features
-
-    fetchResults().then((results) => {
-      console.log(results);
+    fetchResults().then((results: any) => {
```
wu-lee, line 45: Ditto - types will appear here too

apps/front-end/src/components/panel/directoryPanel/DirectoryPanel.tsx
```diff
@@ -40,17 +39,12 @@ const DirectoryPanel = () => {
...
+      if (!results.failed) setResultsTotals(results);
```
wu-lee, line 46: My typechecker says `results` could be undefined here, in which case trying to get `results.failed` may be an exception

wu-lee:
> This seems fragile to me. I've gotten the country code from the item json because it's in the index 1 position on the array. Is this the best way currently available to do this?

I think if you make the suggested change(s) the types work out better.

I haven't followed the logic of where `itemValues` comes from back to the source, but it's an array of strings, or a string. So if you assign it to a local variable, this should at worst get you an undefined. Then TS's checker will be able to infer that the local variable's type correctly as `"string"` you exclude everything but `typeof "string"` (which it can't if it's dealing with a dereferenced array element, for technical reasons).

Then the switch I suggest later should get clean typechecking on the result body.

wu-lee:
Failed to find a simple reference for these points, but the term is "type narrowing" and the point I mean is probably in buried here somewhere https://www.typescriptlang.org/docs/handbook/2/narrowing.html

wu-lee:
> ...from the item json because it's in the index 1 position on the array. Is this the best way currently available to do this?

Actually I just remembered enough about how it works to answer this a bit more.

I think the fields are stored as an array of arrays. But this is for space efficiency reasons in the back end. In the front end, or a single item request from the back end. I'm not certain if this is actually done, and what you say suggests not, but I think perhaps they *ought* to be reconstituted as JSON objects with sensible keys.

So I think if there's no obvious way of doing that you can see, we should have a tech-debt issue requesting that.

PR #143, opened 2025-04-08
Cleanup

wu-lee:
#### What? Why?

Related to issue #95 and the last release

Post-release work. This is a combination of:
- A clean-up of the code-base deferred until after the release
- Steps to automate the release process, which proved to be a bit awkward and likely error prone in future.

- Code should build. Tests should pass
- Clean-up should be uncontroversial.
- Release script needs some inspection, see changes to top-level `package.json` and `bin/do-release`.
- Additional dependency `zx` added to facilitate scripting. 

#### Checklist

- [x] Test-run the release process as per the docs in `docs/versioning-and-releases.md`

The test run should be basically a matter of:

```
npm run release v99.9.9
```

It should result in a new commit tagged `v99.9.9` if successful. Try it again with the same tag, and it should fail leaving a branch `prepare-release-v99.9.9`, which can be manually deleted. (The idea is that the failed process is rolled back, but can be inspected before deletion.)

Check, are there any other failure modes?

This script is meant to save time and errors in future, without sinking too much time on making it perfect - therefore the intention is to  be a best effort and that improving it can be ongoing work.

#### Deployment notes

No. Just remember to delete any tags left from testing so they don't get pushed into the main repository.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4,13 +4,18 @@
   "requires": true,
   "packages": {
     "": {
+      "name": "mykomap-monolith",
       "workspaces": [
         "apps/*",
         "libs/*"
       ],
+      "bin": {
+        "do-release": "bin/do-release"
+      },
       "devDependencies": {
         "editorconfig": "^2.0.0",
-        "rimraf": "^6.0.1"
+        "rimraf": "^6.0.1",
+        "zx": "^8.5.0"
       }
     },
     "apps/back-end": {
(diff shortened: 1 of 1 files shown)


wu-lee:
These all work for me, so will merge - I doubt they will, but if problems arise we can address these as they come up.

wu-lee:
Note, this PR touches a lot of files, but pretty much all changes are cosmetic changes implemented automatically by PrettierJS.

Those which are not:
- .gitignore
- .prettierignore
- bin/do-release
- docs/versioning-and-releases.md
- package.json
- package-lock.json

Issue #142, opened 2025-04-04
[POWYS] Create intial Powys Food Map


lin-d-hop:
## Description

Let's get an initial implementation of the Powys Food Map live.

The data for the map is here.

Other related issues in this epic are:
-  #141 Enable bespoke markers
- #125 Design bespoke markers
- #126 Design data import
- #124 Check if we can hack a nested category

In this work we want to do a manual import of the data. We just want to use a default marker. We are happy to have a flat category structure.

As the other issues in this epic are completed we can update the map. Feel free to incorporate any of the above work that can be, or to wait for any of the above if it makes sense to do so.

## Acceptance Criteria

- [ ] An initial Powys Food Map is live

rogup:
It probably makes sense to work off the MM branch that I started here https://github.com/DigitalCommons/mykomap-monolith/issues/124

And also I made an initial config.json for the Powys map as part of the above spike too.

wu-lee:
Current spike code is now covered by this draft PR: https://github.com/DigitalCommons/mykomap-monolith/pull/146

Data for this spike is on the [`powys-food-map`](https://github.com/DigitalCommons/cwm-test-data/tree/powys-food-map) branch of the now misnamed cwm-test-data, with the dataset ID `powys-spike`.

And code to generate a `standard.csv` data file from the client's spreadsheet referenced in https://github.com/DigitalCommons/mykomap-monolith/issues/126 has been created in `open-data` on the [`powys`](https://github.com/DigitalCommons/open-data/tree/powys) branch. This CSV has not yet deployed on dev.data.digitalcommons.coop, but it has been run though the MM4.x data import to create the above test data, so the above branch has real data.

The relevant branches of the data and MM have been deployed on dev-2, so the data can be viewed here: https://dev.maps.coop/cwm/?datasetId=powys-spike

lin-d-hop:
Great starting point!
I'm going to close this issue and open the next ones for the tweaks and fixes needed.
Thanks @wu-lee

Issue #132, opened 2025-03-26
[EPIC] [Powys]  Powys Food Map


lin-d-hop:
## Acceptance Criteria

- [ ]

Issue #124, opened 2025-03-24
[Powys][SPIKE] Check if we can hack a nested category structure


lin-d-hop:
[Track under Powys Food Map]

## Description

Powys Food are creating a new MM food systems locally. They would like a nested category structure. The structure is:
```
Emergency food provision
- Food bank
- Community fridge
- Pay as you feel meals
- Community food share
Community growing
- Incredible Edible
- Community garden
- Community orchard
- Allotments
Local food provider
- Farmer/grower
- Cafe/restaurant
- Market
- Shop
Events
- Food festival
- Cooking classes
```

The proposed solution is to:
- Create a category list with all of these categories as a flat list
- Use text to create the indentations eg (“Community growing” vs “  -  Allotments”)
- Map points may belong to multiple categories (eg a category and a subcategory)
- In the call we discussed map points being in multiple categories and multiple subcategories, but the clients data format currently doesn’t support that.
- Ensuring the correct category and subcategory is the responsibility of the client.

Here is a sketch of the Directory panel:

![Image](https://github.com/user-attachments/assets/5228b9b6-0064-45c6-90fb-725014b9ca2b)

This is a spike to ensure this is all possible

## Acceptance Criteria

- [ ] Estimate how long this implementation will take in the new MM.

rogup:
@lin-d-hop Just to clarify, you want all the items in that flat list to be clickable, so that e.g. if you click 'Emergency food provision', it will select all Food banks, Community fridges, etc.?
And e.g. 'St. Martin's Church foodbank' would be a member of both the 'Emergency food provision' category and '- Food bank' "fake nested" category?

rogup:
Also, is there anything we need to check for the 'search' panel? e.g. will we need a filter for any of the categories/sub-categories and if so, how would they want this to appear?

lin-d-hop:
Yes please everything clickable ie categories and sub-categories.

The drop down filter can appear in the same way ie with indented sub-categories.

I think they can also appear this way in the tag in the pop ups. This would be the place they might look a little weird, so this would be good to understand.

rogup:
Spike completed

Code here: https://github.com/DigitalCommons/mykomap-monolith/compare/dev...powys-food-map

**Summary:** It's possible to achieve this with minimal code changes. For the code changes that need to be made, I think it will require ~2-3 days of work

Things still to add/fix:

[...]

PR #146, opened 2025-04-30
Powys food map spike

wu-lee:
#### What? Why?

Related to epic #132 (and in particular issues #124 and #142)

This PR contains the spike implementation for the map code.

Build should be error free. Tests should pass.  

#### Checklist

- [x] Added UTs

The map should look like a good working version with the right markers and the right directory and filters when supplied with the latest dataset, called `powys-food-systems`, which needs to be deployed for this to be available.

Searching and filtering should work correctly.

The map is not expected to zoom correctly to the bounding box of the pins.

If some datapoints are in the US, this is Powys' spreadsheet addresses being sketchy.

This PR doesn't aim to get everything perfect, e.g. language support comes later. 

#### Deployment notes

The latest Powys dataset, called `powys-food-systems`, needs to be deployed for this to be available to test against.

i.e. This should work: http://dev.maps.coop/?datasetId=powys-food-systems


rogup:
@wu-lee what's happening with this PR? should it be in progress?

wu-lee:
Ok, the reason why they don't is: the locale is different, so the datestring is not what is expected. Solution: set the locale in `package.json` before node starts running tests via the `LC_ALL` environment variable. Setting it in the test using `vi.stubEnv()` doesn't work, nor does using `process.env`. 

In tracking this down on GitHub CI without access to the temporary files generated, I needed to see the differences in the test output, and so I added this functionality to the tests.

Issue #153, opened 2025-05-14
[POWYS] Add Custom Marker Designs to the Powys Map


ms0ur1s:
### Description
Add the new custom markers to the Powys Food Map. The relevant SVGs are included within the issue's resources along with an icon set guide. The SVGs are already correctly coloured and the guide contains a colour reference for any related styling. 

### Tasks
- [ ] Implement custom markers for the relevant categories
  - Emergency Food Provision
  - Community Growing
  - Local Food Provider
  - Events
- [ ] Check that markers are applied to the correct categories

### Acceptance Criteria
- Custom icons are present on the Powys map
- Icons are attributed to the correct categories
- There are no adverse effects to the map's functionality

***

### Related Epic and Issues
- https://github.com/DigitalCommons/mykomap-monolith/issues/148
- https://github.com/DigitalCommons/mykomap-monolith/issues/125
- https://github.com/DigitalCommons/mykomap-monolith/issues/141

### Icon Resources
- [Powys Custom Marker SVGs](https://github.com/user-attachments/files/20245102/Powys.Custom.Marker.SVGs.zip)
- [Powys Food Map Icon Set Guide](https://github.com/user-attachments/files/20245081/Powys.Food.Map.Icons.Set.pdf)

ms0ur1s:
No worries @King-Mob, I'll get these done this morning.

wu-lee:
The problem with PNGs is they have their colours baked in, and this was a problem with MM3.x, which used them, because it meant there was a fixed palette of colours, and that broke down when datasets needed more because of the categories in use. And you can't really predict what that requirement is.

I'd hoped we could avoid supplying pre-coloured marker icons and use markers which are colour agnostic, that can be tinted dynamically using (say) CSS. 

But I think what this implies is that we can't do that, and (in future when we implement custom colours and icons) we will have to supply N icons in M colours?

ms0ur1s:
[Powys Custom Marker PNGs](https://github.com/user-attachments/files/20366592/PNGs.zip)

King-Mob:
Thanks for pngs, have include in #141

Issue #141, opened 2025-04-04
[POWYS] Enable bespoke markers in MM


lin-d-hop:
## Description

For the Powys Map they would like to use bespoke markers for their categories. As they are exploring nested categories, a single icon will likely apply to multiple categories.

The marker icons are being designed in issue #125 
The initial map is being created in #142 

## Acceptance Criteria

- [x] Outline the approach for implemenation 
- [x] Enable bespoke marker icons to be used for specific categories in MM

## Other notes

This would be a good one to pair on. It would make sense to have Nick in conversations as much thinking has gone into bespoke markers in the past.

Let's not overengineer this. I strongly recommend implementing for the specific use case. If there are most complex things needed in the future we can implement them then.

rogup:
@lin-d-hop / @ms0ur1s What are the final designs for the markers?

lin-d-hop:
Awaiting client feedback

lin-d-hop:
Is this one actually blocked without the final marker designs? Like can you pick a random icon and implement and we just switch in the final designs later?

wu-lee:
Beginning a review - First pass review comments added - based on visual inspection on GH, I've not delved deeper using an editor to check types and things so far.

wu-lee:
I think it needs some more work to deal with potentially multiple categories on a data item, so I am making some suggested changes on the PR.

lin-d-hop:
Ok, thanks Nick. Re the spreadsheet changing - these were your recommendations so I'm surprised this blew out the work. Particularly on this unrelated issue.

When I zoom in on the map I don't see any markers (macos, firefox). So I can't pass the QA on this one at this stage.
dev.maps.coop/cwm/?datasetId=powys-spike

https://github.com/user-attachments/assets/fe5c4de5-9a15-4cf6-b073-c523256c894a

wu-lee:
Use https://dev.maps.coop/cwm/?datasetId=powys-food-systems and it should work.

The spreadsheet changes were recommended, but they still require work to adapt to when implemented.  I've more in fact, suggestions to make their spreadsheet / our map more robust, so it would be good to coordinate on that.

That was an expected change - the unexpected change (as least for me) was needing to cater for items which have *more* than one food category, and having show that situation as a marker.  (Even if they have do the same supercategory, this is a) dependent on them remembering to follow that rule, b) still requires dealing with in the data normalisation, because we are filtering/directorying that category.)

lin-d-hop:
[...]

PR #155, opened 2025-05-20
Bespoke map markers

King-Mob:
#### What? Why?

Related to issue #141

New property added to config.json, e.g. `ui: { marker_property_name: "food_system_category"}`
This picks up a category in the csv and uses it for the markers.

This category looks like `fs:cg` for Community Growers.
In `apps/front-end/src/components/map/markers.ts` the category names are matched to icon files.
The icon files have to be png.

#### Checklist

- [ ] Added UTs

- Connect to a dataset that has marker_property_name in the config
- Look at markers

#### Deployment notes

No.

### DIFF
--- a/apps/front-end/src/components/map/mapLibre.ts
+++ b/apps/front-end/src/components/map/mapLibre.ts
@@ -186,10 +186,9 @@ export const createMap = (
     const markerList = [];
     let index = 0;
 
-    let markerName: keyof typeof markers;
-    for (markerName in markers) {
-      const markerImage = markers[markerName];
+    for (let markerImage of markers) {
       const image = await map.loadImage(markerImage);
+      const markerName = "marker-" + index;
       map.addImage(markerName, image.data);
       markerList.push(index++);
       markerList.push(markerName);
--- a/apps/front-end/src/components/map/markers.ts
+++ b/apps/front-end/src/components/map/markers.ts
@@ -4,14 +4,34 @@ import emergencyFoodMarker from "./markers/emergency-food-marker.png";
 import eventsMarker from "./markers/events-marker.png";
 import localFoodMarker from "./markers/local-food-marker.png";
 
-//when adding new markers, ensure default is still the last entry
-//these labels are hardwired and this is just a temporary situation, to be rectified later
-const markers = {
-  "fs:cg": communityGrowingMarker,
-  "fs:ef": emergencyFoodMarker,
-  "fs:ev": eventsMarker,
-  "fs:lm": localFoodMarker,
-  "default-marker": defaultMarker,
-};
+// When adding new markers, ensure default is still the last entry. FIXME These labels
+// are hardwired and this is just a temporary situation, to be rectified later.
+// See comments here:
+// https://github.com/DigitalCommons/mykomap-monolith/pull/155#discussion_r2144547335
+//
+// The order should match the definition of terms in the vocab property named by
+// config.json's ui.marker_property_name value. There should be one extra icon
+// for cases which take multple values, if these cases exist.
+const markers = [
+  emergencyFoodMarker, // "ep": "Emergency food provision",
+  emergencyFoodMarker, // "ep-fb": " - Food bank",
+  emergencyFoodMarker, // "ep-cf": " - Community fridge",
+  emergencyFoodMarker, // "ep-pm": " - Pay as you feel meals",
+  emergencyFoodMarker, // "ep-cf": " - Community food share",
+  communityGrowingMarker, // "cg": "Community growing",
+  communityGrowingMarker, // "cg-ie": " - Incredible Edible",
+  communityGrowingMarker, // "cg-ga": " - Community garden",
+  communityGrowingMarker, // "cg-or": " - Community orchard",
+  communityGrowingMarker, // "cg-al": " - Allotments",
+  localFoodMarker, // "lp": "Local food provider",
+  localFoodMarker, // "lp-gr": " - Farmer/grower",
+  localFoodMarker, // "lp-cr": " - Cafe/restaurant",
+  localFoodMarker, // "lp-mk": " - Market",
+  localFoodMarker, // "lp-sp": " - Shop",
+  eventsMarker, // "ev": "Events",
+  eventsMarker, // "ev-ff": " - Food festival",
+  eventsMarker, // "ev-cc": " - Cooking classes"
+  defaultMarker, // No multi-category case anticipated
+];
 
 export default markers;


wu-lee:
I've changed the base branch to `powys-food-map` since earlier commits are outside the scope of the parent issue.

apps/back-end/src/dataset.ts
```diff
@@ -162,8 +164,8 @@ export class DatasetWriter {
       // they stand out as different. The result is intentionally somewhat CSV-like.
       searchableFile.write(
         `{   "itemProps":\n` +
-          JSON.stringify(filterablePropNames) +
```
wu-lee, line 165: Whitespace change - not a problem in itself but it looks like your editor is not applying the PrettierJS formatting styles used in the project. We agreed to use this back at the start of the project to eliminate "stylistic churn" - exactly this formatting change which depends on the idiosyncrasies of an individual's editor config or personal style.

Suggest addition of the changes from "npm run prettier:format" at the end of the PR to work around that, and addition of the PrettierJS plug-in for VSCode. I think you can turn it on in a per-project way but I'm not an expert, @rogup might know better.

apps/back-end/package.json
```diff
@@ -12,7 +12,7 @@
   },
   "scripts": {
     "clean": "rimraf node_modules dist node-dist",
-    "dataset": "tsx bin/dataset.ts",
+    "dataset": "vite-node bin/dataset.ts",
```
wu-lee, line 15: curious what which motivated this change?

apps/back-end/src/cli/dataset.ts
```diff
@@ -69,6 +69,7 @@ export class ImportCmd extends Command {
     // Reason: we don't want anything silently stripped, and probabaly nor do we
     // want to silently pass through extras.
     const config = schemas.ConfigData.parse(await slurpJson(this.configPath));
+    console.log(config)
```
wu-lee, line 72: Looks like a debug trace. Is this deliberate and intended to be permanent?

apps/front-end/src/components/map/mapLibre.ts
```diff
@@ -157,19 +157,33 @@ export const createMap = (
...
     map.addLayer({
       id: "unclustered-point",
       type: "symbol",
       source: "items-geojson",
       filter: ["!", ["has", "point_count"]],
-      layout: {
-        "icon-image": "default-marker",
-        "icon-offset": [0, -20], // shift marker icon up so tip is at the marker's coordinates
-      },
+      layout: markerLayout as any
```
wu-lee, line 186: Is using `any` avoidable here?

apps/front-end/src/components/map/markers.ts
```diff
@@ -0,0 +1,16 @@
+import defaultMarker from "./markers/map-marker.png";
+import communityGrowingMarker from "./markers/community-growing-marker.png";
+import emergencyFoodMarker from "./markers/emergency-food-marker.png";
+import eventsMarker from "./markers/events-marker.png";
+import localFoodMarker from "./markers/local-food-marker.png";
+
+//when adding new markers, ensure default is still the last entry
+const markers = {
```
wu-lee, line 8: Might be worth leaving a big notice in a comment that these labels are hardwired and this is just a temporary situation, to be rectified later.

apps/front-end/src/components/popup/leftPane/LeftPane.tsx
```diff
@@ -10,7 +10,7 @@ import { renderIfData } from "../../../utils/jsx-utils";
 
 interface LeftPaneProps {
   name: string;
-  food_system_category: string[];
+  food_system_category?: string;
```
wu-lee: Likewise, big warning comment saying this is very Powys-specific and should be unhardwired later would be good.

libs/common/src/api/mykomap-openapi.json
```diff
@@ -868,6 +868,9 @@
                       "properties": {
                         "directory_panel_field": {
                           "type": "string"
+                        },
+                        "marker_property_name": {
+                          "type": "string"
```
wu-lee, line 873: I find commits to this file, being generated automatically, want to be in their own commits - like those to `package-lock.json`. Otherwise later on they have a tendency to tangle up with other logical changes when merging/rebasing etc.

apps/front-end/src/components/popup/leftPane/LeftPane.tsx
wu-lee, line 1: Looks like this change introduced type errors in the tests? I see "Unchanged files with check annotations" errors from GH's validation below.

apps/back-end/package.json
King-Mob: the import from the front-end into the back-end doesn't work with tsx, because it rejects files with a .png extension

apps/front-end/src/components/map/mapLibre.ts
King-Mob: not really, I can get you slightly closer but it still wants me to use an unknown before converting to the type it's using internally. I think the issue is the lisp-like syntax that the type system isn't parsing

apps/back-end/src/cli/dataset.ts
```diff
@@ -87,7 +88,7 @@ export class ImportCmd extends Command {
 
     try {
       const dsWriter = new DatasetWriter(propDefs);
-      const stats = await dsWriter.writeDataset(this.dataPath, "id", csvReader);
+      const stats = await dsWriter.writeDataset(this.dataPath, "id", csvReader, config.ui.marker_property_name);
```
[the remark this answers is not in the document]
King-Mob: it's consistent with the property name that was above it in config.ui

King-Mob:
@wu-lee moved the files around, let me know what you think

King-Mob:
This also closes #153

King-Mob:
Added testing notes.

I don't know if the Powys dataset has been generated yet. @wu-lee @rogup ?

wu-lee:
I've rebased this branch onto the end of `dev` and made minor amendments to the expected cases in `searchSlice.test.ts`, to work in the new context, which has changed as a result of 534f77b1a80d49b5378e9 on PR #167.

Builds and tests ok.

rogup (changes requested):
It's looking mostly good, I've left a few comments.

Also, I think we need to add some marker icons to the dataset-cli test data. And also to add them to the test `dataset-A` data, so we have a reference in the codebase for how the config needs to be set up to make them work.

apps/back-end/src/cli/dataset.ts
```diff
@@ -86,7 +88,96 @@ export class ImportCmd extends Command {
...
+          override markerIndex(item: DatasetItem): number | undefined {
+            let value = item[markerName];
+
+            // Check there's a value present. Note the loose inequality,
+            // matches null or undefined.
+            if (value == undefined) return undefined;
+
+            if (markerPropDef.type !== "multi") {
+              // A singlar value
+              return terms.indexOf(String(value));
```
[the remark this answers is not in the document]
wu-lee: I see your point.

I think the way it works here is just intended to be a stop-gap, clearly it's all very hardwired to the Powys case. We could have more discussion about this, but we need to decide how much we want to do right now in this PR, and perhaps this is just adequate?

rogup:
Also, I reviewed this since it was 'In Review' and I saw I had been requested as a reviewer on the PR. Looks like it was John who did this.

I think we need to change our process since this happens quite a lot, where a review is done when it didn't need to be since it was sitting in 'In Review'. I propose we change 'In Review' to 'Waiting for Review', for tickets that are not currently being looked at. Then either
a) move things to 'In Progress' when a review starts
b) have separate 'Waiting for Review' and 'In Review' columns

I think I prefer b), since it will be consistent with our general workflow of moving tickets left to right. Let's bring this to the next delivery meeting

apps/back-end/src/cli/dataset.ts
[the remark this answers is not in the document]
wu-lee: One extra thought revisiting this: the order of the indexing used for markers is currently *defined from* the order terms in the `config.json` definition (for the first AKA default language). So if are always using the `config.json` order I think this will keep things consistent.

What we will need later is a way to map these index numbers to arbitrary icons, replacing this array. That will probably look like a look-up from indexes or vocab term IDs in the `config.json`, which indeed should be validated as self-consistent on loading it.  Although we will probably also need to be able to define what to use in the "default" case (no value/unrecognised value) and the "multiple" case (more than one value).

wu-lee:
This is now merged into the base `powys-merge-map` branch.

Issue #164, opened 2025-05-28
[POWYS] Add new popup layout field to the ConfigData schema


rogup:
## Description

Following on from #149 , the next task for generalising the popup code is to add the the new popup layout config field to the `ConfigData` schema in `contract.ts`.

The design of the new popup config field is laid out in https://github.com/DigitalCommons/mykomap-monolith/blob/powys-food-map/docs/config.md#popup-layout

Then, once added to the `contract.ts`, implement this new field in both the cwm-latest and powys-spike config.json on this branch https://github.com/DigitalCommons/cwm-test-data/tree/powys-food-map/datasets

## Acceptance Criteria

- [ ] The new design for the popup config is implemented in the contract.ts
- [ ] The new popup config is added to both the CWM and Powys map config

ColmDC:
Only insignificant whitespace differences between checked layouts in https://dev.maps.coop/cwm/?datasetId=cwm-latest and https://maps.coop/cwm/?datasetId=cwm-latest

Is there anything else I could check?

PR #207, opened 2025-07-31
add popup to config contract

King-Mob:
#### What? Why?

Related to issue #164

#### Checklist

- [ ] Added UTs

- Click '...'

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -29,13 +29,15 @@ And these are the different options to define the style of an itemProp:
 - `singleColumnLimit` (number): If specified, a list will split into 2 columns if the number of values is more than this.
 - `showLabel` (defaults to `false`): If this is set to true, the label of the itemProp (taken from `titleUri` in the itemProp definition) is displayed as a heading above the value(s).
 - `hyperlinkBaseUri` (defaults to None): If `valueStyle` is `hyperlink`, this base URI is prepended to the itemProp value, to form the href.
+- `displayText` (defaults to None): If `valueStyle` is `hyperlink`, this sets the display text of the hyperlink.
 
 ```
 "popup": {
     "titleProp": "name",
     "left-pane": [
       {"itemProp": "category", "valueStyle": "text", "showLabel": true },
-      {"itemProp": "description", "valueStyle": "text" }
+      {"itemProp": "description", "valueStyle": "text" },
+      {"itemProp": "website", "valueStyle": "hyperlink", "displayText": "Website" }
     ],
     "top-right-pane": [
       {"itemProp": "address", "valueStyle": "address" },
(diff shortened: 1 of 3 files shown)


libs/common/src/api/contract.ts
```diff
@@ -107,6 +107,15 @@ const PropSpec = z.discriminatedUnion("type", [
   MultiPropSpec,
 ]);
 const PropSpecs = z.record(z.string(), PropSpec);
+const popupItem = z.object({
+  itemProp: z.string(),
+  valueStyle: z.union([z.literal("text"), z.literal("address"), z.literal("hyperlink")]).default("text"),
```
rogup, line 112: this can be optional too, not sure if that makes a difference in TS validation since it's maybe implied by default

libs/common/src/api/contract.ts
```diff
@@ -107,6 +107,15 @@ const PropSpec = z.discriminatedUnion("type", [
...
+  showBullets: z.boolean().default(false).optional(),
+  singleColumnLimit: z.number().optional(),
+  showLabel: z.boolean().default(false).optional(),
+  hyperlinkBaseUri: z.string().default("").optional(),
+  displayText: z.string().default("").optional()
```
rogup, line 117: I feel like this probably shouldn't have a default value, since we don't want it to default to the empty string. We just want to show the raw URL if displayText isn't supplied

docs/config.md
```diff
@@ -29,13 +29,15 @@ And these are the different options to define the style of an itemProp:
...
 - `hyperlinkBaseUri` (defaults to None): If `valueStyle` is `hyperlink`, this base URI is prepended to the itemProp value, to form the href.
+- `displayText` (defaults to None): If `valueStyle` is `hyperlink`, this sets the display text of the hyperlink.
 
 ```
 "popup": {
     "titleProp": "name",
     "left-pane": [
       {"itemProp": "category", "valueStyle": "text", "showLabel": true },
-      {"itemProp": "description", "valueStyle": "text" }
+      {"itemProp": "description", "valueStyle": "text" },
```
rogup, line 39: Now that we have a zod schema, I think we should use this as the source of truth and link to it from the docs, since the docs will get out of date unless we remember to update them after every change.

We can do this by adding the descriptions of each field to the Zod schema with `.openapi({ description: ''})`

Also, the openapi function has a `example` field so we could include the whole example from the docs too.

rogup (commented):
Hey @King-Mob 
this is looking good, I've included some very minor comments + a suggestion for the docs

docs/config.md
King-Mob: This seems like a bigger task, I'll make an issue for it.

Issue #165, opened 2025-05-28
[POWYS] Implement front-end generalised popup code


rogup:
## Description

After completing #164 , the next task is to implement the generalised popup code in the front-end that converts the popup layout in the config.json into MUI components. Work on the `powys-food-map` branch

The design of the new popup config field is laid out in https://github.com/DigitalCommons/mykomap-monolith/blob/powys-food-map/docs/config.md#popup-layout

During development, test with the cwm-latest and powys-spike datasets that contain the new popup config in https://github.com/DigitalCommons/cwm-test-data/tree/powys-food-map/datasets

## Acceptance Criteria

- [ ] There is generalised popup code in the front-end that doesn't contain any hardcoded names of CWM or Powys itemProps. There is no visible difference in appearance or function of the CWM popups.

PR #208, opened 2025-08-02
165 generalised popup code front end

King-Mob:
#### What? Why?

Related to issue #165

#### Checklist

- [ ] Added UTs

- Open the popups and check they render correctly
- Banco Creditcoop is a good one to check

### DIFF
--- a/apps/front-end/src/app/configSlice.ts
+++ b/apps/front-end/src/app/configSlice.ts
@@ -22,6 +22,24 @@ export type ConfigLogo = {
   };
 };
 
+export type PopupItem = {
+  itemProp: string;
+  valueStyle: "text" | "address" | "hyperlink";
+  showBullets?: boolean;
+  singleColumnLimit?: number;
+  showLabel?: boolean;
+  hyperlinkBaseUri?: string;
+  displayText?: string;
+  multiple?: boolean;
+};
+
+export type ConfigPopup = {
+  titleProp: string;
+  leftPane: PopupItem[];
+  topRightPane: PopupItem[];
+  bottomRightPane: PopupItem[];
+};
+
 export interface ConfigSliceState {
   vocabs: Config["vocabs"];
   languages: string[];
@@ -29,6 +47,7 @@ export interface ConfigSliceState {
   map?: ConfigMap;
   logo?: ConfigLogo;
   status: "idle" | "loading" | "loaded" | "failed";
+  popup?: ConfigPopup;
 }
 
 const initialState: ConfigSliceState = {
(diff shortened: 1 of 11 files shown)


apps/front-end/src/app/configSlice.ts
```diff
@@ -22,13 +22,32 @@ export type ConfigLogo = {
   };
 };
 
+export type PopupItem = {
```
rogup, line 25: from my code read of the latter files, I think PopupItemConfig will be clearer

apps/front-end/src/components/popup/PopupItems.tsx
```diff
@@ -0,0 +1,210 @@
...
+import List from "@mui/material/List";
+import ListItem from "@mui/material/ListItem";
+import Link from "@mui/material/Link";
+
+const removeHtmlTags = (text: string) =>
+  decode(
+    text.replace(/<[^<>]+(>)/g, ""), // remove HTML tags
+  )?.trim();
+
+const Text = ({ item, text }: { item: PopupItem; text: string }) => {
```
rogup, line 16: Renaming `item` to `itemConfig` makes more sense when I'm reading this

apps/front-end/src/components/popup/popupSlice.ts
```diff
@@ -15,16 +14,7 @@ interface PopupSliceState {
   // TODO: handle different fields rather than hardcoding them, use config to define how they should
```
rogup, line 14: can remove this comment

docs/config.md
```diff
@@ -29,19 +29,22 @@ And these are the different options to define the style of an itemProp:
 - `singleColumnLimit` (number): If specified, a list will split into 2 columns if the number of values is more than this.
 - `showLabel` (defaults to `false`): If this is set to true, the label of the itemProp (taken from `titleUri` in the itemProp definition) is displayed as a heading above the value(s).
 - `hyperlinkBaseUri` (defaults to None): If `valueStyle` is `hyperlink`, this base URI is prepended to the itemProp value, to form the href.
+- `displayText` (defaults to None): If `valueStyle` is `hyperlink`, this sets the display text of the hyperlink.
+- `multiple` (defaults to false): If this is true, the value is an array.
```
rogup, line 33: I don't think we need this new state. It can be inferred from the config.itemProps since we already specify here if a value is multi. You might need to work out how to pass this through the redux state

rogup (commented):
@King-Mob From a quick code read, this is looking really nice, good job!

My main bit of feedback is that an example of this new config should be added to the test dataset-A config.json. This will allow me to run the code myself on my dev environment and then I'll give it a second pass of the review where I can see what's happening. How were you testing it up till now?

Also, I think we should deffo schedule work to introduce automated visual regression tests. This is la creme de la creme of features for such tests, where we can feed in config + data and expect a component to appear exactly how we want. @lin-d-hop is this something that we can use the SABI tech debt budget for? If it fits into our tech debt priorities, I can't remember where we recorded those.

rogup:
Added to https://github.com/DigitalCommons/technology-and-infrastructure/issues/131

Issue #204, opened 2025-07-16
[CWM]  [POWYS] Englarge powys popup right pane to 50% for longer emails


King-Mob:
## Description

Sometimes the powys email links are too long for the popup box panel

## Acceptance Criteria

- On powys maps, the popup right pane is 50% of the popup

rogup:
@King-Mob Could I propose that instead of doing this, we change emails to just use a mailto: link with link text "Email"? Like we do for the website field. This feels nicer to me from a UX perspective, and would be more robust since someone could have an email that's still too long.

PR #213, opened 2025-08-12
make the width configurable

King-Mob:
#### What? Why?

Related to issue #204

Add config value `leftPaneWidth` to popup config, with a default value of 70%

#### Checklist

- [ ] Added UTs

- Open powys-eng and notice the luxurious extra space for email address
- Open cwm-latest and notice that the popups look the same as they ever did


rogup (commented):
Looks good, just 1 comment

Issue #159, opened 2025-05-22
[SABI] Sync pop up open state into URL params


ms0ur1s:
### Description
When a pop up is opened via `openPopup` thunk in `popupSlice.ts`, the URL must be updated to include `?popupId=<ID>` using React Router’s shallow routing. If needed, perform routing in a component callback.

### Tasks

- [x]  Research React Router's shallow routing API
- [x] Update `openPopup` thunk to add pop ID to URL params
- [x] If thunk can’t access router, create a wrapper component that listens for `openPopup` actions and updates URL
- [x] Verify opening pop up updates URL correctly
- [x] Test back / forward browsers navigation retains pop up state

### Acceptance Criteria

- Dispatching `openPopup` appends `popupId to current URL
- Back/forward browser button maintains the correct pop up state
- Manual URL entry with `?popupId=` opens the correct pop up

***

### Related Issues
- https://github.com/DigitalCommons/mykomap-monolith/issues/158
- https://github.com/DigitalCommons/mykomap-monolith/issues/160
- https://github.com/DigitalCommons/mykomap-monolith/issues/161
- https://github.com/DigitalCommons/mykomap-monolith/issues/162

### Related Epic and Spike
- https://github.com/DigitalCommons/technology-and-infrastructure/issues/128
- https://github.com/DigitalCommons/mykomap-monolith/issues/144

### Resources
- https://medium.com/@codingscenes/shallow-routing-is-a-feature-provided-by-react-router-8d134e209a1d
- https://stackoverflow.com/questions/70928255/what-is-the-best-way-to-use-react-router-v6-navigation-with-redux-and-redux-thun/73449716#73449716

King-Mob:
@ColmDC  Currently the zoom hasn't been specified to be a url parameter, but for a digital commons popup id to be open it would look like this:

`https://maps.coop/cwm/?datasetId=cwm-latest&popupId=demo%2Fplus%2Fdc%2Fl5yy6l`

King-Mob:
@ColmDC The id is generated by our data merge system

rogup:
Testing details for QA: https://github.com/DigitalCommons/mykomap-monolith/pull/249

rogup:
@ColmDC One thing that's dawned on me is that if the app is being used through an iframe or fullscreen on a mobile device, it won't be obvious that these links can be shared. Should we make a ticket to add a share button to the UI that copies the current URL to the clipboard?

ColmDC:
[...]

PR #214, opened 2025-08-13
159 put popup open state in url

King-Mob:
#### What? Why?

Related to issue #159

Add react-router to handle searchparams in react components

#### Checklist

- [ ] Added UTs

- Click on a marker, notice that it's id is in the url
- Click on the map to close popup, notice url now gone
- Click on marker, then refresh the page with the popupid in the url, the popup will open


apps/front-end/src/components/map/mapLibre.ts
```diff
@@ -340,14 +341,14 @@ export const createMap = (
       if (e.features) {
         const feature = e.features[0] as GeoJSON.Feature<GeoJSON.Point>;
         const coordinates = feature.geometry.coordinates.slice();
-        const itemIx = feature.properties?.ix;
+        const itemId = feature.properties ? `@${feature.properties.ix}` : "-1";
 
-        if (popup?.isOpen() && popupIx === itemIx) {
+        if (popup?.isOpen() && popupId === itemId) {
```
rogup, line 346: Did this work in your testing? I don't think it will, since we're trying to compare IDs with Indexes.

The reason why mapLibre.ts must use indexes is because the first data that the FE requests (the `locations` route) only contains a list of coordinates. This is to minimise the size of data needed to render the initial map, since we're working with hundreds of thousands of markers. Therefore, mapLibre only knows about indexes (the ix property in each marker feature) and not IDs. If we want to change the way that mapLibre is displaying e.g. apply a filter, we must supply it a list of indexes.

I realise that even though this was a key part of our design choice and we spent a lot of time thinking about it, we haven't properly documented this. I'm going to make a tech debt issue to rectify this.

What this means for this PR is that changes from index to id should be reverted, and we should feed in indexes when calling the `openPopup` mapLibre event.

apps/front-end/src/components/map/MapWrapper.tsx
```diff
@@ -117,6 +127,14 @@ const MapWrapper = () => {
     map.current?.fire("changeLanguage", { language });
   }, [language]);
 
+  useEffect(() => {
+    const id = searchParams.get("popupId");
+    if (id && mapCreated) {
+      dispatch(openPopup(id));
+    }
+
+  }, [searchParams, mapCreated])
```
rogup, line 136: This should just depend on `mapCreated`, not `searchParams` too. Generally we want to keep the architecture of Redux being the source of truth and the components (and searchParams) to reflect Redux, rather than the other way around. However, this is solving the edge case where a shareable link has been used and the map has just been created but Redux hasn't 'caught up' yet, so we need to dispatch a Redux action. It would be good to also add a comment to explain that this is its one purpose

I think in its current form this `useEffect` may be covering up the break in mapLibre's functionality

apps/front-end/src/components/map/MapWrapper.tsx
```diff
@@ -99,7 +109,7 @@ const MapWrapper = () => {
       console.log("Opening popup");
       if (popupLocation) {
         map?.current?.fire("openPopup", {
-          itemIx: popupIndex,
+          itemId: popupId,
```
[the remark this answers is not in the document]
rogup: Also this `useEffect` is a good place to add the logic to also keep the URL params in sync with the Redux state. There could be a line at the top of the useEffect to do this.

PR #237, opened 2025-11-24
159 put popup open state in url

### LINKED ISSUE #159: [SABI] Sync pop up open state into URL params

King-Mob:
#### What? Why?

Fixes #159

#### Checklist

- [ ] Added UTs

- open popup, look at url
- close popup, look at url
- open popup, refresh, see url, see popup

### DIFF
--- a/apps/back-end/src/routes.ts
+++ b/apps/back-end/src/routes.ts
@@ -7,8 +7,6 @@ import {
   getDatasetConfig,
   getDatasetItemByIx,
   getDatasetItemById,
-  getDatasetItemByIx,
-  getDatasetItemById,
   getDatasetLocations,
   getTotals,
   initDatasets,
--- a/apps/back-end/src/routes.ts
+++ b/apps/back-end/src/routes.ts
@@ -101,8 +101,6 @@ export function MykomapRouter(
       const itemId = itemIdOrIx;
       const item = getDatasetItemById(datasetId, itemId);
 
-      console.log("ITEM ITEM ITEM", item)
-
       return { status: 200, body: item };
     },
 
--- a/apps/back-end/src/routes.ts
+++ b/apps/back-end/src/routes.ts
@@ -103,6 +103,8 @@ export function MykomapRouter(
       const itemId = itemIdOrIx;
       const item = getDatasetItemById(datasetId, itemId);
 
+      console.log("ITEM ITEM ITEM", item)
+
       return { status: 200, body: item };
     },
 
--- a/apps/back-end/src/routes.ts
+++ b/apps/back-end/src/routes.ts
@@ -7,6 +7,8 @@ import {
   getDatasetConfig,
   getDatasetItemByIx,
   getDatasetItemById,
+  getDatasetItemByIx,
+  getDatasetItemById,
   getDatasetLocations,
   getTotals,
   initDatasets,
@@ -52,7 +54,7 @@ export function MykomapRouter(
   if (!fs.existsSync(opts.mykomap.dataRoot))
     throw new Error(
       `the dataRoot plugin option is set but refers to a non-existing path: ` +
-        `'${opts.mykomap.dataRoot}'.`,
+      `'${opts.mykomap.dataRoot}'.`,
     );
 
   console.log("Initialising datasets...");
@@ -103,7 +105,7 @@ export function MykomapRouter(
       const itemId = itemIdOrIx;
       const item = getDatasetItemById(datasetId, itemId);
 
-      return { status: 200, body: item };
+      return { status: 200, body: { ...item } };
     },
 
     getTotals: async ({ params: { datasetId } }) => {
--- a/apps/back-end/src/routes.ts
+++ b/apps/back-end/src/routes.ts
@@ -105,7 +105,7 @@ export function MykomapRouter(
       const itemId = itemIdOrIx;
       const item = getDatasetItemById(datasetId, itemId);
 
-      return { status: 200, body: { ...item } };
+      return { status: 200, body: item };
     },
 
     getTotals: async ({ params: { datasetId } }) => {
--- a/apps/back-end/src/routes.ts
+++ b/apps/back-end/src/routes.ts
@@ -101,6 +101,8 @@ export function MykomapRouter(
       const itemId = itemIdOrIx;
       const item = getDatasetItemById(datasetId, itemId);
 
+      console.log("ITEM ITEM ITEM", item)
+
       return { status: 200, body: item };
     },
 
--- a/apps/back-end/src/routes.ts
+++ b/apps/back-end/src/routes.ts
@@ -91,9 +91,7 @@ export function MykomapRouter(
     },
 
     getDatasetItem: async ({ params: { datasetId, datasetItemIdOrIx } }) => {
-      const itemIdOrIx = Buffer.from(datasetItemIdOrIx, "base64").toString(
-        "utf8",
-      );
+      const itemIdOrIx = Buffer.from(datasetItemIdOrIx, 'base64').toString('utf8');
 
       if (itemIdOrIx.startsWith("@")) {
         const itemIx = Number(itemIdOrIx.substring(1));
--- a/apps/back-end/src/routes.ts
+++ b/apps/back-end/src/routes.ts
@@ -7,6 +7,8 @@ import {
   getDatasetConfig,
   getDatasetItemByIx,
   getDatasetItemById,
+  getDatasetItemByIx,
+  getDatasetItemById,
   getDatasetLocations,
   getTotals,
   initDatasets,
--- a/apps/back-end/src/routes.ts
+++ b/apps/back-end/src/routes.ts
@@ -5,10 +5,8 @@ import fs from "node:fs";
 import {
   getDatasetAbout,
   getDatasetConfig,
-  getDatasetItemByIx,
   getDatasetItemById,
   getDatasetItemByIx,
-  getDatasetItemById,
   getDatasetLocations,
   getTotals,
   initDatasets,
@@ -54,7 +52,7 @@ export function MykomapRouter(
   if (!fs.existsSync(opts.mykomap.dataRoot))
     throw new Error(
       `the dataRoot plugin option is set but refers to a non-existing path: ` +
-      `'${opts.mykomap.dataRoot}'.`,
+        `'${opts.mykomap.dataRoot}'.`,
     );
 
   console.log("Initialising datasets...");
@@ -91,19 +89,19 @@ export function MykomapRouter(
     },
 
     getDatasetItem: async ({ params: { datasetId, datasetItemIdOrIx } }) => {
-      const itemIdOrIx = Buffer.from(datasetItemIdOrIx, 'base64').toString('utf8');
+      const itemIdOrIx = Buffer.from(datasetItemIdOrIx, "base64").toString(
+        "utf8",
+      );
 
       if (itemIdOrIx.startsWith("@")) {
         const itemIx = Number(itemIdOrIx.substring(1));
         const item = getDatasetItemByIx(datasetId, itemIx);
 
-        return { status: 200, body: { ...item, index: `@${itemIx}` } };
+        return { status: 200, body: { ...item, index: itemIx } };
       }
 
       const itemId = itemIdOrIx;
       const item = getDatasetItemById(datasetId, itemId);
-
-      console.log("ITEM ITEM ITEM", item)
 
       return { status: 200, body: item };
     },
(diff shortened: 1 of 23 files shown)


apps/back-end/src/services/Dataset.ts
```diff
@@ -101,33 +101,27 @@ export class Dataset {
...
 
   getItemById = (itemId: string) => {
-    console.log("itemid", itemId);
     // Use searchable if it has ids, otherwise loop through items
     if (this.searchablePropIndexMap.id) {
-      const itemIx = this.searchablePropValues.findIndex(
-        (item) => item[this.searchablePropIndexMap.id] == itemId,
-      );
-      if (itemIx > -1) {
+      const itemIx = this.searchablePropValues.findIndex(item => item.includes(itemId));
```
rogup: This isbold code that I left a comment about a while ago and has been fixed in the previous PR that went into dev. This change is reverting it

apps/back-end/src/services/Dataset.ts
```diff
@@ -101,33 +101,27 @@ export class Dataset {
...
+      if (itemIx) {
         const item = this.getItemByIx(itemIx);
-        return { ...item, index: `@${itemIx}` };
+        return { ...item, itemIx, index: `@${itemIx}` }
```
rogup: We don't need this new itemIx field. The index is already there in the 'index' field

apps/back-end/src/services/Dataset.ts
```diff
@@ -101,33 +101,27 @@ export class Dataset {
...
-        itemIx < this.searchablePropValues.length;
-        itemIx++
-      ) {
+    }
+    else {
+      for (let itemIx = 0; itemIx < this.searchablePropValues.length; itemIx++) {
         const item = this.getItemByIx(itemIx);
         if (item.id === itemId) {
-          return { ...item, index: `@${itemIx}` };
+          return { ...item, itemIx, index: `@${itemIx}` };
```
rogup: ditto

apps/back-end/src/routes.ts
```diff
@@ -89,9 +89,7 @@ export function MykomapRouter(
     },
 
     getDatasetItem: async ({ params: { datasetId, datasetItemIdOrIx } }) => {
-      const itemIdOrIx = Buffer.from(datasetItemIdOrIx, "base64").toString(
-        "utf8",
-      );
+      const itemIdOrIx = Buffer.from(datasetItemIdOrIx, 'base64').toString('utf8');
```
rogup: Are you sure your IDE is correctly set up to use the editor config file? I think it's maybe being overridden by your local user settings, causing lots of churn in indentation that makes things difficult to review (and makes git blame history confusing too)

apps/back-end/test/plugin.test.ts
```diff
@@ -263,8 +263,8 @@ const encodeBase64 = (data: string) => {
   return Buffer.from(data).toString("base64");
 };
 
-describe("getDatasetItem", () => {
-  describe("dataset exists", () => {
+describe("dataset exists", () => {
```
rogup: I think these naming changes might be a mistake

apps/front-end/src/components/map/mapLibre.ts
```diff
@@ -348,14 +354,14 @@ export const createMap = (
       if (e.features) {
         const feature = e.features[0] as GeoJSON.Feature<GeoJSON.Point>;
         const coordinates = feature.geometry.coordinates.slice();
-        const itemIx = feature.properties?.ix;
+        const itemIx = feature.properties ? `@${feature.properties.ix}` : "-1";
 
-        if (popup?.isOpen() && popupIx === itemIx) {
+        if (popup?.isOpen() && popupId === itemIx) {
```
rogup: comparing ID to index won't work. Maplibre.ts should only deal in indexes, not IDs. 
Did you see my comments in the previous PR? https://github.com/DigitalCommons/mykomap-monolith/pull/214

apps/front-end/src/components/map/MapWrapper.tsx
```diff
@@ -93,13 +102,37 @@ const MapWrapper = () => {
...
+      setLoaded(true);
+    }
     // Keep the mapLibre popup in sync with the Redux state
     if (popupIsOpen) {
       console.log("Opening popup");
+      tryUpdateSearchParams(popupId);
       if (popupLocation) {
         map?.current?.fire("openPopup", {
-          itemIx: popupIndex,
+          itemIx: `@${popupIndex}`,
```
rogup: I think it's useful in this instance to keep index as a number and not add the @, so we can differentiate between IDs and indexes with types, to ensure we don't accidentally pass an id to maplibre.ts

apps/front-end/src/components/map/MapWrapper.tsx
```diff
@@ -93,13 +102,37 @@ const MapWrapper = () => {
     }
   }, [features, sourceLoaded]);
 
+  function tryUpdateSearchParams(popupId: string) {
```
rogup: why 'try' and not just updateSearchParams? there isn't a reason why it would fail

apps/front-end/src/components/map/MapWrapper.tsx
```diff
@@ -93,13 +102,37 @@ const MapWrapper = () => {
...
+  function tryDeleteSearchParams() {
+    const urlPopupId = searchParams.get("popupId");
+    if (urlPopupId) {
+      searchParams.delete("popupId");
+      setSearchParams(searchParams);
+    }
+  }
+
   useEffect(() => {
+    if (!loaded && mapCreated) {
```
rogup: I think it's simpler and makes things easier to read if this block is in a separate useEffect that just depends on mapCreated. Then you won't need the 'loaded' state since it will only trigger once when the map is created.

(Reasoning is that I personally avoid adding too many state variables like 'loaded' unless they're absolutely necessary, since they can create tangles of logic and race conditions that often lead to bugs in the future. In this case, we do need the mapCreated state, so we can trigger stuff with the URL on first load, so we don't need loaded too)

apps/front-end/src/components/map/MapWrapper.tsx
```diff
@@ -108,10 +141,22 @@ const MapWrapper = () => {
         // This is handled in Popup.tsx
       }
     } else {
+      if (loaded) {
```
rogup: This could be if(mapCreated) so that loaded isn't needed

apps/front-end/src/components/map/MapWrapper.tsx
```diff
@@ -108,10 +141,22 @@ const MapWrapper = () => {
...
+        tryDeleteSearchParams();
+      }
       console.log("Closing popup");
       map?.current?.fire("closeAllPopups");
     }
-  }, [popupIsOpen, popupIndex]);
+  }, [popupIsOpen, popupIndex, mapCreated]);
+
+  useEffect(() => {
```
rogup: I don't think we want this useEffect. Redux should be our source of truth and we make searchParams a mirror of this (not vice versa, except in the case of first load, which is handled above). This is also explained in https://github.com/DigitalCommons/mykomap-monolith/pull/214

apps/front-end/src/components/panel/resultsPanel/results/Results.tsx
```diff
@@ -52,12 +53,17 @@ const Results = () => {
...
+    new window.URLSearchParams(),
+  );
 
   const isMedium = useMediaQuery("(min-width: 897px)");
 
   const onItemClick = (itemIx: number) => {
     console.log(`Clicked item @${itemIx}`);
-    dispatch(openPopup(itemIx));
+    dispatch(openPopup(`@${itemIx}`));
+    // searchParams.set("popupId", `@${itemIx}`);
```
rogup: remove comments

apps/front-end/src/components/popup/popupSlice.test.ts
```diff
@@ -102,6 +108,7 @@ describe<LocalTestContext>("popup reducer", (it) => {
       id: "test/cuk/R000003",
       name: "Another Cooperative Organization",
       primary_activity: "aci:ICA220",
+      itemIx: 7
```
rogup: the API doesn't use itemIx: 7, it uses index: '@7'

apps/front-end/src/components/popup/popupSlice.test.ts
```diff
@@ -37,7 +43,7 @@ describe<LocalTestContext>("popup reducer", (it) => {
   });
 
   it("should handle openPopup andclosePopup", async ({ store }) => {
-    await store.dispatch(openPopup(0));
+    await store.dispatch(openPopup('@0'));
```
rogup: Since openPopup now handles opening by ID too, we should add a test case for opening by ID

apps/front-end/src/components/map/MapWrapper.tsx
```diff
@@ -25,17 +27,23 @@ const MapWrapper = () => {
...
+  const [searchParams, setSearchParams] = useSearchParams(
+    new window.URLSearchParams(),
+  );
 
-  const popupCreatedCallback = (itemIx: number) => {
-    console.log("Popup created");
-    dispatch(openPopup(itemIx));
+  const popupCreatedCallback = (ix: string) => {
+    console.log("Popup created", ix);
+    dispatch(openPopup(ix));
```
rogup: This should now be openPopup(`@${ix}`)

apps/front-end/src/components/map/mapLibre.ts
```diff
@@ -131,11 +136,12 @@ const onMarkerHover = (
  * Set up the sources and layers of the MapLibreGL map instance.
  */
 export const createMap = (
-  popupCreatedCallback: (itemIx: number) => void,
+  popupCreatedCallback: (id: string) => void,
```
rogup: This should still be ix. Maplibre.ts only deals in indexes

apps/front-end/src/components/map/mapLibre.ts
```diff
@@ -232,14 +238,14 @@ export const createMap = (
         leafOffset: [number, number],
       ) => {
         const coordinates = feature.geometry.coordinates.slice();
-        const itemIx = feature.properties?.ix;
+        const itemIx = `@${feature.properties?.ix}`;
 
-        if (popup?.isOpen() && popupIx === itemIx) {
+        if (popup?.isOpen() && popupId === itemIx) {
```
rogup: keep as popupIx

apps/front-end/src/components/map/mapLibre.ts
```diff
@@ -22,7 +22,8 @@ export const POPUP_CONTAINER_ID = "popup-container";
 const POPUP_INITIAL_ZOOM = 15;
 
 let popup: Popup | undefined;
-let popupIx: number | undefined;
+let popupId: string | undefined;
```
rogup: Don't add this new popupId variable to this file and keep popupIx as a number. Maybe also add a note at the top saying that we only deal with indexes in this file with a link to the architecture doc which talks about this, for future devs. Since it's quite confusing and an easy mistake to make.

rogup (changes requested):
Hj @King-Mob ik there are a lot of comments but this is a complicated PR and vast majority of it is looking very good. Most of comments are about some id / index confusion still, and also an architectural thing about source of truth + useEffect loops, which hopefully I've explained ok!

PR #245, opened 2025-12-02
[front-end] open results panel on mobile if results exist

### LINKED ISSUE #185: [CWM]  Can't open search results on mobile after search
**Describe the bug**
On mobile, after you have conducted a search and selected from the search results, you can't return to those search results.

**To Reproduce**

1. Open https://dev.maps.coop/cwm/?datasetId=cwm-latest on a mobile screen
2. Open search 
3. Search for text, 'app' in the text box
4. Select the Agriculture Primary activity and then Apply filter
4. Select the last page of results
5. Select _TAPPAL SAHAKARI SANGH_ from the results
6. You decide you want to choose a different result form the last page of results.
7. There is no quick way to get back to search results panel.
8. If you select search again, you need to reapply the search and navigate again to the last page of results 

**Expected behaviour**
It should be possible to return to that results page more easily

**Additional context**

ms0ur1s:
#### What? Why?

Fixes #185 

After performing a search and selecting a result on mobile, you cannot return to the search results without reproducing the search. I have updated the logic in `Panel.tsx`, so that if a search has been performed, tapping the search tab will return you to the results panel.

* Added a check in the tab selection handler to automatically open the results panel on mobile (`!isMedium`) when the user selects tab 2, a filter is active, and there are visible search results.
* Integrated the `selectVisibleIndexes` selector into the component to determine if there are visible results after filtering. [[1]](diffhunk://#diff-375ae61a737aaef08b2827fdcb7afad7692cdc743a072dde77347cf8ac2856c0R27) [[2]](diffhunk://#diff-375ae61a737aaef08b2827fdcb7afad7692cdc743a072dde77347cf8ac2856c0R65-R70)

- Perform a search on mobile
- Tap on a result
- Tap on the search tab
- Search results pane should be visible
- Clear the search results
- Tap on a tab other than the search tab
- Return to the search page
- The filters page should be visible
- Perform a search that returns 0 matches
- Tap on a tab other than the search tab
- Return to the search pane
- The filters page rather than the results should be displayed

### DIFF
--- a/apps/front-end/src/components/panel/Panel.tsx
+++ b/apps/front-end/src/components/panel/Panel.tsx
@@ -24,6 +24,7 @@ import {
 import {
   performSearch,
   selectIsFilterActive,
+  selectVisibleIndexes,
 } from "../panel/searchPanel/searchSlice";
 import { useTranslation } from "react-i18next";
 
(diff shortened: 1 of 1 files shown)


rogup (approved):
Looks good!

PR #247, opened 2025-12-09
180 Pan to filter results

### LINKED ISSUE #180: [CWM] Pan to cluster centre  when country is selected
### Description
When a country is selected either within the directory or the search filter, the map should pan to the centre of that country's marker clusters. 

### Acceptance criteria
- When a country is selected the map pans to the centre of the country's marker clusters

***

#### Nice to have, but not necessary
As both left-hand panels will be expanded at this point, it would be nice if:
- the country's marker clusters centred horizontally and vertically within the map's remaining visible space
- The map zoomed in enough to give the country increased prominence
<br/>

Image

rogup:
#### What? Why?

Fixes #180 

As I outline in this [comment](https://github.com/DigitalCommons/mykomap-monolith/issues/180#issuecomment-3629860548), this PR adds panning/zooming so that the results of any filter (including clicking on a directory field) are centered on the map and zoomed to an appropriate level.

The issue also said that it wants the behaviour to account for whether the left panel(s) are open when positioning the map. I have achieved this by syncing the Redux panel open state to mapLibre.ts and doing some basic calculations to calculate the offset in pixels.
After implementing this, I noticed some inconsistency since when we pan to markers in other situations (e.g. when clicking on them or clicking on a search result) we don't account for the panels. So I improved the UX in these cases too, so we always take panels into account.

The bulk of the changes are in mapLibre.ts since this is where we want to have the MapLibre related logic such as panning and zooming. The only other changes are to MapWrapper, to trigger the panning to happen when a filter is applied, and to sync the panel open state since maplibre.ts doesn't have direct access to Redux (as mentioned above).

I used Claude code to implement this fix using these 2 prompts:

> Implement this feature: https://github.com/DigitalComm ons/mykomap-monolith/issues/18 0#issuecomment-3629860548 , specifically the solution proposed in the comment I linked. In summary, when a filter is applied, either by selecting a directory category or using the search filter, the map should pan and zoom in to the results so that these markers are central on the screen, and zoomed in to fill the screen. This feature only requires changes to the front-end app. It may be possible to implement it entirely in the mapLibre.ts file, by using a map listener for when new data is set

> Add a bit of padding to the bounds so that the points are not right on the edge of the screen. Also, on desktop when the left panel and the resultsPanel are visible after a search, many of the points are behind these panels. Shift the position of the map so that the markers are shifted to the right and more of them are visible to the right of the panels

I then spent roughly an hour manually refactoring what it had written and making a few more changes. Maybe I could have achieved these further changes with more prompts, but I would have had to written a lot of words to describe exactly what I wanted and I don't yet trust that it would have done what I wanted.

#### Checklist

- No UTs needed. Changes were only to front-end maplibre logic, which UTs are not appropriate for

The best way to test these changes is to do some general smoke testing of filtering on both desktop and mobile. On desktop, try filtering/clicking on markers with the left panels both expanded and collapsed.


King-Mob (approved):
Looks good. This isn't for this issue but I really want to integrate mapLibre into the react tree more. 

Maybe if we could somehow get rid of all the event listeners and only use functions that would be a good first step?

PR #248, opened 2025-12-10
228 Hide the results panel instead of destroying it so scroll state remains

### LINKED ISSUE #228: [CWM]  Search result scroll position discarded after selection
**Describe the bug**
After selecting a co-op from the search results, the scroll position in the search results is lost. When there are more than a page worth of co-ops per page this can be very frustrating.

Add/fix ability to recall scroll position in search results panel.
 

**To Reproduce**

1. Go to https://maps.coop/cwm/?datasetId=cwm-latest
2. In the Directory select a country with >100 co-ops.
3. Scroll down to the bottom of the resultant list and select the bottom co-op
4. Reopen the search panel and you will be back at the top of the list 

**Expected behaviour**
The result panel should display the same co-ops it was displaying when closed until the search criteria get changed.

rogup:
#### What? Why?

Fixes #228 

Not much more detail is needed, it's a small fix

I initially used Claude code to tackle this and it tried to use a fairly obscure external library 'use-scroll-restoration' hook to try to track and restore scroll position. But then it tried to add more complicated layers of useEffect logic to get it to work whilst also being able to reset the scroll position when the results page changed. It might have reached an extremely complicated workaround eventually, but I eventually had to manually intervene and revert its edits and do this fix myself.

#### Checklist

- No docs needed
- No UTs needed

### DIFF
--- a/apps/front-end/src/components/panel/Panel.tsx
+++ b/apps/front-end/src/components/panel/Panel.tsx
@@ -67,10 +67,8 @@ const Panel = () => {
 
   const isMedium = useMediaQuery("(min-width: 897px)");
 
-
   const handleToggle = () => {
     dispatch(togglePanel());
-    console.log("isOpen", isOpen);
   };
 
   const handleTabChange = (tab: number) => {
@@ -91,7 +89,6 @@ const Panel = () => {
   const handlePanelClose = () => {
     dispatch(closePanel());
     if (!isMedium) dispatch(setSelectedTab(0));
-    console.log("isOpen", isOpen);
   };
 
   const handleMapTapClick = () => {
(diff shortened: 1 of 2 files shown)


rogup:
@ms0ur1s Mabybe try reviewing this. The changes look more complicated than they are because there's some indentation change, but there's actually basically only 1 or 2 lines that have changed.

ms0ur1s (approved):
Cool, thanks @rogup  for passing this my way to review. It all makes sense to me and I can see what you've done. You're no longer unmounting `StyledResultsPanel` and removing it from the DOM, instead you're using the drawer's `open` prop to hide the component via the styled wrapper, and preserving the scroll state.

PR #252, opened 2025-12-28
175 Add analytics to MM

### LINKED ISSUE #175: [SABI] Add analytics to MM
## Description

Following our meeting about analytics (notes [here](https://docs.google.com/document/d/1gWSr70pVpyPYJ7Sn188MONZxP_zi3ZIvtpHTtjngXsc/edit?tab=t.0), the most relevant part is at the bottom of the doc), we now need to implement the first round of analytics in MM.

We are following [Mixpanel's advice](https://docs.mixpanel.com/docs/tracking-best-practices/tracking-plan) and have created a MM Tracking Plan here: 
https://docs.google.com/spreadsheets/d/1lVnfCVoc0Nj_C5bqvL2VtfXvLtJKkY0bmq120EUjxlM/edit?gid=1641519238#gid=1641519238

We need to implement all of the events in the Tracking Plan on the MM front-end. All the interaction with Mixpanel should occur in a contained part of the code e.g. in an `analytics.ts` file, exposing a `trackEvent` function that can be used elsewhere in the code, so that we can switch analytics providers or move the implementation to the server-side if we decide to do this in the future.
This code from the LX Mixpanel spike may be useful: https://github.com/DigitalCommons/land-explorer-front-end/compare/development...349-user-analytics

## Acceptance Criteria

- [x] All events in the Tracking Plan are implemented
- [x] The Tracking Plan is updated with all code-specific details and implemented analytics are ticked off
- [x] Manual verification that the analytics are firing and showing up in the Mixpanel dashboard

rogup:
#### What? Why?

Fixes #175 

See https://docs.google.com/spreadsheets/d/1lVnfCVoc0Nj_C5bqvL2VtfXvLtJKkY0bmq120EUjxlM/edit?gid=1641519238#gid=1641519238 for explanation of changes

It will be easiest to review each commit separately in order. Everything's quite self-explanatory apart from the commit that simplifies the configSlice, which isn't directly related to this analytics work. I just ended up fixing it up since configSlice was getting unwieldy and it was throwing type errors when I made a little change to the contract. The changes should avoid this happening again.

#### Checklist

- not applicable for front-end stuff like this

- We should test that each of the analytics in the tracking plan is firing as expected and visible in the Mixpanel console
- And that there aren't false positives. e.g. a 'share' analytic shouldn't be fired when you refresh the app or re-open a link within the same browser. It should only fire if a link is shared to a different browser

#### Deployment notes

When deploying to dev-2/prod-2, we need to set up a project on the Mixpanel console if one hasn't been set up already and put the appropriate token into the front-end's .env


apps/front-end/src/components/map/mapLibre.ts
King-Mob, line 1: do we need more logs? already quite noisy

apps/front-end/src/services/types.ts
```diff
@@ -2,3 +2,20 @@ import { ClientInferResponseBody } from "@ts-rest/core";
 import { contract } from "@mykomap/common";
 
 export type Config = ClientInferResponseBody<typeof contract.getConfig, 200>;
+
+// This is the config of a popup item as definted in a dataset config.json
```
King-Mob: *defined?

King-Mob (approved):

PR #253, opened 2025-12-30
244 cwm add a dotcoop variant map marker 2

### LINKED ISSUE #244: [CWM] Add a DotCoop Variant Map Marker
## Description

Add a new map marker variant with a for co-ops that use .coop domains.

### Tasks

- [x] Add logic to select this marker if `.coop` domain is present
- [x] Verify correct behaviour at all zoom levels
- [ ] Add key if necessary

### Acceptance Criteria
- DotCoop Markers appear only for organisations with a .coop domain
- Works seamlessly with clustering and pop-up triggers.

### Reference

Image

ms0ur1s:
### What? Why?
To add support for a DotCoop marker that visually distinguishes organisations with a DotCoop datasource.

Fixes #244 

### Code-specific details (for Reviewer)

**Front-end marker logic and assets:**

* Added a new `dotCoopMarker` icon and assigned it to the index corresponding to "DC" (DotCoop verified) in the `markers` array, while most other data sources use the `defaultMarker`. Powys specific markers (community growing, emergency food, events, local food) are commented out. [[1]](diffhunk://#diff-761f830489c1131a15c5742eba001c9e6203e4ff74ede50312c7c3498735aa56L2-R6) [[2]](diffhunk://#diff-761f830489c1131a15c5742eba001c9e6203e4ff74ede50312c7c3498735aa56R16-R61)
* Updated comments and index mapping in `markers.ts` to match the CWM data source vocabulary.

**Back-end marker selection logic:**

* Fixed the logic in the dataset import command so that for single-value arrays, the correct index is used (`values[0]` instead of `values[1]`), and for multiple-value arrays, the presence of "DC" will now prioritize showing the DotCoop marker.

- Zoom to a location (US has a good mix)
  - Check for instances of the DotCoop marker
    - Check these instances contain a DotCoop datasource
  - Check the datasources of a couple of default markers (ensuring there's no DotCoop)
  - Check an organisation with multiple datasources that include DotCoop (Green Fields Market)
    - Ensure it has a DotCoop marker
  - Check another org with multiple datasources that doesn't include Dotcoop
    - Should display default marker
- Check a spiderfied location
  - Search for: NATIONAL COOPERATIVE DAIRY FEDERATION OF INDIA LIM
  - Visit location
  - Clear search
  - Click on the cluster when in appears in that location
  - Check a DotCoop marker is present
- Have a general play with the map


rogup:
Hey @ms0ur1s the new design looks good. But the issue is that this will break the Powys maps, due to the hardcoding of its markers being commented out. All our maps use the same client code. To achieve with hardcoding, you would have to keep the Powys hardcoded icon values and add the new DC marker icon to the list, and use this new index in the CWM for verified co-ops.

However, I think this actually shows that the hardcoding is confusing and an easy source of bugs, so IMO we should just invest now, as part of this issue, to make the icons configurable via config without hardcoding. I can think of a way that this could be achieved fairly easily. Would you prefer me to:
- briefly describe my idea for making the icons fully configurable and have a go yourself
- take on this task, and have a call afterwards where I can explain what I did. Maybe we could pair for part of this, but would need to think about the best way to do this, since we weren't that productive last time we tried pairing
?

ms0ur1s:
Hey @rogup, thank you for explaining this. Your proposal makes great sense.

As far as how we progress, I'm happy either way. Although as much as I like the idea of option one, option two might be better, minus the pairing and you just explaining how it works, afterwards

rogup:
I believe @King-Mob is working on this currently so will assign this ticket to him

wu-lee:
Just noting that this probably needs to be rebased onto a recent version of `dev` before reviewing...

PR #262, opened 2026-01-20
202 /  201 CWM introduce badge for co-located clusters

### LINKED ISSUE #201: [CWM]  Improve interaction: auto-zoom colocated clusters
## Description
Clusters that contain only colocated markers currently require multiple zoom interactions to reach their final location. This can feel repetitive and unnecessary, as no further declustering is required.

Clicking a cluster composed solely of colocated markers should trigger a smooth, progressive zoom that takes the user directly to the correct destination in a single interaction, without being visually jarring.

### UX Considerations
When interacting with colocated clusters, users often need to repeat zoom actions to reach the appropriate level, often requiring multiple double-clicks. If no further declustering is needed, the map should determine the correct zoom level and transition there in one smooth action.

## Acceptance Criteria

- Clicking a cluster that contains only colocated markers triggers a single zoom interaction
- The map smoothly transitions to the correct zoom level for the colocated markers
- No additional manual zoom actions are required to reach the final marker state
- The zoom behaviour is not abrupt or visually jarring
- Clusters containing non-colocated markers retain existing declustering behaviour

<br>

### LINKED ISSUE #202: [CWM]  Design Visual cue for cluster marker with only colocated entities
## Description
There is no way of identifying if a marker cluster contains only colocated entities. Design a visual cue that identifies when this is the case. Reflect on if behavavior change is alo required.

This will be helpful as the spider view of clustered markers only triggers when you are at the deepest zoom level. Single clicking on a cluster that has non colocated zooms a level and declusters a level. Single click does nothing if there is no declustering left to do. So perhaps this needs to also be accompanied by the behaviour change to zoom a level if you single click on cluster that includes only co-located co-ops? Or a zoom right in to selected co-op option? 

## Acceptance Criteria
- [ ] Selection of designs/colour changes for clusters that contain only co-located entities
- [ ] Make supporting behaviour change recommendations to support

ms0ur1s:
### What? Why?
Sometimes multiple organisations are registered at the same location (co-located). When zoomed in these entities are represented by spiderfied markers, however, when zoomed out they are indistinguishable from any other cluster. As a consequence, it is often necessary to click and zoom numerous times before reaching the point where the cluster spiderfies. This can be frustrating as the user is forced into a repetative task without any indication of progress.

This enhancement adds a visual indicator, in the form of a badge, to the top right hand corner of co-located clusters. The badge is activated when a co-located cluster is clicked. In addition, co-located clusters now zoom three levels when clicked, reducing the journey to the end location.

Fixes #201 / #202 

### Code-specific details (for Reviewer)

**Co-located cluster badge feature:**

* Added a new SVG badge icon and circle layer (`colocated-badge-circle` and `colocated-badge-icon`) to visually indicate co-located clusters on the map, with logic to load and display the badge only for relevant clusters.
* Implemented the `updateBadgeVisibility` function to dynamically show or hide badges based on which clusters are co-located and whether they are currently spiderfied.

**Co-located cluster detection and interaction:**

* Added logic to detect when all points in a cluster are co-located, track these cluster IDs in a set, and update badge visibility accordingly during cluster click events.
* Modified cluster click behavior to zoom in by 3 levels for co-located clusters, helping users reach the spiderfy threshold more quickly, and to update the spiderfied cluster state.

**Event handling improvements:**
* Ensured badge visibility and spiderfied state are reset appropriately on map zoom events.

- Navigate to India
- Click on a large cluster
- At each zoom level click on the largest cluster
  - You should eventually hit a co-located cluster 
  - A co-located cluster badge should appear
  - You should zoom three level at a time
- Click until you reach maximum zoom
  - When the cluster spiderfies the badge should disappear
- Zoom out and the badge should re-appear
  - It should disappear when the cluster envelopes another marker / cluster
- Navigate to another large cluster and repeat

<br/>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -32,5 +32,6 @@ site
 
 # MS local test files
 ResultsPanel copy 2.tsx
+apps/back-end/test/data/datasets/
 
 bin/do-release.mjs
--- a/apps/front-end/src/components/map/mapLibre.ts
+++ b/apps/front-end/src/components/map/mapLibre.ts
@@ -21,10 +21,11 @@ export const POPUP_CONTAINER_ID = "popup-container";
 /** The level to which we zoom when jumping to a popup, on clicking a search result */
 const POPUP_INITIAL_ZOOM = 15;
 
-/** Note: We ONLY USE INDEXES in this file. MapLibre doesn't know about IDs. Read architecture.md */
-let popupIx: number | undefined;
 let popup: Popup | undefined;
+let popupIx: number | undefined;
 let tooltip: Popup | undefined;
+let colocatedClusterIds = new Set<number>();
+let currentSpiderfiedClusterId: number | null = null;
 
 /**
  * We need to offset latitude of the map centre slightly above a marker's location when opening a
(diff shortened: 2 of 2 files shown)


King-Mob (changes requested):
Please could the svg and the code loading it and preparing it to be used be moved into it's own file. 

My ideal would be that you end up importing a function, and then using it in this file by doing something like `const badgeImage = await loadBadge()` and then it gets added to the map..

Other than that, it looks good. github says there's a merge conflict to resolve, so if you can do that as well that'd be great.

If one or both of these turns into a >30 minute headache please let me know

PR #274, opened 2026-02-27
261 [CWM] Update Primary CTA Colour

### LINKED ISSUE #120: [CWM] Update Primary CTA Colour
### Description
Users testing suggested that the primary call-to-action (CTA) buttons are not immediately noticeable, especially the Apply Filters button and the tab that opens the side panel. The current green colour blends with the UI, making key actions less prominent.

**Suggested Solutio**n
- Added a new colourway to the CWM palette 
- Update the key CTAs with this colour

***

**Notes**
No. 12 on the 'Suggested Improvements' list in issue #104 is worth considering in relation to this.

### LINKED ISSUE #261: [CWM] Update Primary CTA Colour
## Description
In response to the resolution of issue #120. Update the MM CTA colour to match the related [Figma designs](https://www.figma.com/design/xCGRZZgXqCVkM5AoBhpuln/MM--CWM-UI-Improvements?node-id=19-1621&t=ydx5zc2PqseSeIZd-1).

### Tasks
- Confirm colour variant 
- Apply updated colourway to:
  - [x]  **Apply Filters** button in mobile search pane
  - [x]  **Clear Search** button in the results pane

### Acceptance Criteria
Updated CTA colour:
- Matches the chosen variant
- Enhances the CTA's visibility
- Does not clash with existing colours

<br/>

ms0ur1s:
### What? Why?
Refined the primary call-to-action (CTA) buttons to improve visibility based on user feedback, while maintaining the MykoMap palette and overall UI consistency.

Fixes #261

**Component and API updates:**

* Added an optional `intent` prop (with value "cta") to the `StandardButton` component's interface and passed it through to the underlying MUI `Button` component. (`StandardButton.tsx`) [[1]](diffhunk://#diff-d1276d0143573fb06ed36f76c340471a7542915a589426b4ff4daeca805e5870R7) [[2]](diffhunk://#diff-d1276d0143573fb06ed36f76c340471a7542915a589426b4ff4daeca805e5870R19-R26)

**Usage updates:**

* Updated usages of `StandardButton` in `ApplyFilters` and `ResultsPanel` to use `intent="cta"` for primary action buttons, ensuring consistent visual emphasis for key actions. (`ApplyFilters.tsx`, `ResultsPanel.tsx`) [[1]](diffhunk://#diff-ad00004172893b2458508882fe610d048189e43ee3b41e5aae06eeb339933aa7L55-R55) [[2]](diffhunk://#diff-459962da47b2fc20915ddecc08d3785c6939687c1f97bd667d42236b0ee4da2dL120-R123)

**Theming and styling:**

* Added a custom variant for `Button` components with `variant="contained"` and `intent="cta"`, applying a gradient background, custom hover, focus, and disabled styles in the theme. (`theme.ts`)
* Introduced a new CSS variable `--color-primary-rgb` for use in theme styling. (`GlobalCSSVariables.tsx`)

**Desktop**

- Click on the **Directory** tab
- Click on an entry
- Check the **Clear Search** button:
  - Aligns with the current MM palette
  - Has increased prominence
  - Hover state behaves as expected

 
**Mobile**

- Tap on the **Search** tab
- Select a filter
- Check the **Apply Filters** button:
  - Uses the updated CTA styling
  - Displays correctly in default and pressed states

<br>

### DIFF
--- a/apps/front-end/src/components/common/standardButton/StandardButton.tsx
+++ b/apps/front-end/src/components/common/standardButton/StandardButton.tsx
@@ -4,6 +4,7 @@ import Button from "@mui/material/Button";
 interface ButtonProps {
   variant?: "contained" | "outlined" | "text";
   color?: "primary" | "secondary";
+  intent?: "cta";
   children: React.ReactNode;
   disabled?: boolean;
   buttonAction?: () => void;
@@ -15,12 +16,14 @@ const StandardButton = ({
   children,
   disabled = false,
   buttonAction,
+  intent,
   ...props
 }: ButtonProps) => {
   return (
     <Button
       variant={variant}
       color={color}
+      intent={intent}
       onClick={buttonAction}
       disabled={disabled}
       {...props}
--- a/apps/front-end/src/components/panel/applyFilters/ApplyFilters.tsx
+++ b/apps/front-end/src/components/panel/applyFilters/ApplyFilters.tsx
@@ -52,7 +52,7 @@ const ApplyFilters = ({
         </StandardButton>
       </div>
       <div>
-        <StandardButton buttonAction={buttonAction} disabled={disabled}>
+        <StandardButton intent="cta" buttonAction={buttonAction} disabled={disabled}>
           {buttonText}
         </StandardButton>
       </div>
--- a/apps/front-end/src/components/panel/resultsPanel/ResultsPanel.tsx
+++ b/apps/front-end/src/components/panel/resultsPanel/ResultsPanel.tsx
@@ -117,7 +117,10 @@ const ResultsPanel = () => {
         }}
       >
         <StyledButtonContainer>
-          <StandardButton buttonAction={handleClearSearch}>
+          <StandardButton
+            intent="cta"
+            buttonAction={handleClearSearch}
+          >
             {t("clear_search")}
           </StandardButton>
           {!isMedium && <CloseButton buttonAction={handlePanelClose} />}
--- a/apps/front-end/src/components/panel/resultsPanel/results/Results.tsx
+++ b/apps/front-end/src/components/panel/resultsPanel/results/Results.tsx
@@ -87,6 +87,7 @@ const Results = () => {
         variant="h4"
         component="h4"
         sx={{
+          marginTop: "var(--spacing-medium)",
           color: "var(--color-neutral-tint)",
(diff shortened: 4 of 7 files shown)


apps/front-end/src/components/common/standardButton/StandardButton.tsx
```diff
@@ -4,6 +4,7 @@ import Button from "@mui/material/Button";
 interface ButtonProps {
   variant?: "contained" | "outlined" | "text";
   color?: "primary" | "secondary";
+  intent?: "cta";
```
laurathackray21, line 7: What's the difference between this new `intent` prop and the existing `color` one? I'm wondering why we can't use the color prop for this? (I think color=primary is usually used as the call to action version of a button.)

By introducing a third property for `intent` I think we might risk causing some confusion with things like what a button with color=secondary and intent=cta looks like.

King-Mob:
@ms0ur1s yes, agree with Laura and wondering if we need three colours why we can't add "cta" along with "primary" and "secondary" in the `colour` prop?

apps/front-end/src/components/common/standardButton/StandardButton.tsx
ms0ur1s: Thanks @laurathackray21 and @King-Mob, I had similar thoughts while implementing this.

The idea behind using `intent` was to separate what the button’s action is (e.g. a primary action / CTA) from how it’s coloured (primary / secondary), while also allowing for future action types. `color="primary"` relates to the MM palette and is used throughout the app, which felt a bit restrictive when trying to create a button that stands out from that base colour. My aim was to keep the CTA styling separate from the palette as things evolve.

However, I agree that combining `color="secondary"` + `intent="cta"` could be confusing. So far I’ve only used `intent="cta"` with `color="primary"`, so we could either valid combinations (e.g. CTA always uses / builds upon primary) that or document it more clearly.

Happy to revisit if this feels like overkill.

ms0ur1s: Also, apologies for the late reply. I mean to reply earlier but it slipped my mind.

ms0ur1s: @laurathackray21, I'm going to resolve this for now. We can always revisit later, if needs be.

PR #275, opened 2026-03-04
244 marker icon configurability

### LINKED ISSUE #244: [CWM] Add a DotCoop Variant Map Marker

King-Mob:
#### What? Why?

Fixes #244

Update the data writer so that custom markers come from the config file.

Design doc: https://docs.google.com/document/d/1T8w0Kw3rpIVfUik8lzUeST8JvvnOqr-UrdNbJLMkM-U/edit?tab=t.0

#### Checklist

- [ ] Added UTs

- Click '...'


ms0ur1s:
@King-Mob, dotcoop marker added.

wu-lee:
I've just pushed some tweaks I added earlier when I began reviewing this, hopefully the commit descriptions explain adequately what they are. I think they're still useful to include.

However, since I was requested to defer the reviewing to someone more familiar, happy to leave the choice of whether / where to include these commits to the reviewer/requester.

libs/common/src/api/contract.ts
```diff
@@ -142,6 +142,8 @@ const ConfigData = z.object({
   ui: z.object({
     directory_panel_field: z.string(),
     marker_property_name: z.string().optional(),
```
rogup: I think all these marker fields shouldn't be optional but should be nested within an optional 'ui.customMarkers' config object. Since we do need all the fields to be defined if we are to use custom markers

apps/back-end/src/cli/dataset.ts
rogup: Maybe here it would be good to also validate that the highest index in this map is not greater than the length of the `markerIcons` array

rogup: This case of single value properties (probably the most common case) needs to use the termsToIconIndex too

libs/common/src/api/contract.ts
```diff
@@ -142,6 +142,8 @@ const ConfigData = z.object({
...
+    markerIcons: z.array(z.string()).optional(),
+    termsToIconIndex: z.record(z.number()).optional(),
```
rogup: We should include here that there must always be a 'default' term

libs/common/src/api/contract.ts
rogup: Let's also add some light documentation for the options in config.md to give an example of how they should be used

apps/back-end/src/cli/dataset.ts
```diff
@@ -211,14 +211,27 @@ export class ImportCmd extends Command {
 
         // Must be a list of values
         const values = value as Array<string>;
+        const termsToIconIndex = config.ui.termsToIconIndex;
+
+        if (!termsToIconIndex) return undefined;
```
rogup: If the suggestion of nesting of the options in the contract.ts is implemented, checks like this shouldn't be needed

apps/back-end/src/cli/dataset.ts
```diff
@@ -211,14 +211,27 @@ export class ImportCmd extends Command {
...
+
+        for (let value of values) {
+          if (termsToIconIndex[value] !== undefined) {
+            return termsToIconIndex[value];
+          }
+        }
+
+        return termsToIconIndex["default"];
+
+        /*
```
rogup: Can get rid of this old implementation

apps/front-end/src/components/map/markers.ts
```diff
@@ -3,35 +3,15 @@ import communityGrowingMarker from "./markers/community-growing-marker.png";
 import emergencyFoodMarker from "./markers/emergency-food-marker.png";
```
rogup: We can remove this file and import the markers directly in mapLibre.ts

apps/front-end/src/components/map/mapLibre.ts
rogup: I think here, rather than having to plumb each of the images through markers.ts (which adds more steps every time we want to change/add to the list of marker files), we can simply fetch the image directly using the `markerIcons` list from the config.
e.g. for `marker-1`, the image will be at `./markers/markerIcons[1]` and so on, where markerIcons[1] is something like `dotcoop-marker.png`

PR #277, opened 2026-03-14
Export dataset cli

wu-lee:
#### What? Why?

This PR allows the MM dataset import script to be run outside of this repo, in a node module which adds MM as a dependency, in a fairly minimal way.

The reason for this is to allow other applications, such as the CWM application in the `data-pipeline` project, to build MM data from CSV files.

Without this, the script can only really be run from within a check-out of the MM repository, which makes all sorts of deployment automation awkward or impossible, since the CSV files, config.json and about.md files needed as inputs to the script typically live outside the MM repo, and are concerns that the MM application codebase should not know about, their release process being necessarily independent of MM's release process.

The `data-pipelines` repository currently uses this branch to deploy MM data. After it is merged, it will need to be updated to use dev, or better (when it includes these changes) main. Later this functionality should probably be moved elsewhere closer to the MM data deployment process. This is being worked on in https://github.com/DigitalCommons/technology-and-infrastructure/issues/135

The existing MM codebase should continue to work as before.  Suggest that a build, test and a run are performed to check this.

The data-pipelines codebase should also continue to work. Suggest a build and a run should check this. Note that the action in  that repository already does that, so it simply re-running that and checking it is picking up the right branch of MM to do the data merging may be adequate. However, running it in a check-out may be easier to inspect.

Specifically: the MM dataset schema generated should be the version generated on this PR branch, not an older version.

#### Checklist

- [ ] Dataset formats are correct (see above)

There is probably no QA step required, if the data is correct. Running the data in a MM deployment with the right version would be required to QA it the usual way, and we currently can't do this easily without deploying the data on dev, once dev has this branch merged in.

[EDIT] In deploying the data for [dotcoop-project#80](https://github.com/DigitalCommons/dotcoop-project/issues/80#issuecomment-4061137627), I have effectively QA'ed this works, at least insofar as that:
- the MM data was rebuilt, having rebased `export-dataset-cli` onto the version we want to target, `dev` (this branch needs to be merged for that not to be necessary in future)
- that data was then deployed for dev.maps.coop as `cwm-20260314`, a MM instance which uses the appropriate version of MM (`dev`)
- the dataset was QA'ed visually as both working (at all!) and containing a the expected fix

#### Deployment notes

none


apps/back-end/package.json
```diff
@@ -8,15 +8,17 @@
...
+    "dataset": "dist/dataset"
   },
   "scripts": {
     "clean": "rimraf node_modules dist node-dist",
     "dataset": "vite-node bin/dataset.ts",
     "dataset-minimal": "vite-node bin/datasetMinimal.ts",
     "dataset-minimal-cym": "vite-node bin/datasetMinimalCym.ts",
     "prebuild": "npm -w ../../libs/common -w ../../libs/node-utils run build",
-    "build": "vite build --ssr src/index.ts --outDir dist",
+    "build:lib": "vite build --ssr --outDir dist",
```
rogup, line 23: I think we should start documenting what all these different npm scripts do. It's not obvious just from reading the commands. What's the best practice for this?
Maybe just implement it for this package.json and then we can replicate it in other libs where necessary

rogup (commented):
Looks good, just 1 comment about documentation

apps/back-end/package.json
wu-lee: The reason for renaming to build:lib is explained in the commit message, I assume you saw that, but I take your point.

The JSON format explicitly disallows comments, so it's problematic: inserting comments within the file tends to require hacks with drawbacks (there's even on I found which uses a whole extra npm dependency which allows use of a .JS file), and doing it outside, makes it less visible and accessible. 

The least worst workaround I came across might be to use something like: 

```
    "#prebuild": "description for prebuild...",
    "prebuild": "npm -w ../../libs/common -w ../../libs/node-utils run build",
    "#build": "description for  build...",
    "build": "vite build --ssr src/index.ts --outDir dist",
    "#build:lib": "description for build:lib....",
    "build:lib": "vite build --ssr --outDir dist",
```

(I'm using a hashtag prefix, because that's a shell comment and ensures the command doesn't parse without escaping; the example I saw used JS style `//` comments, although really what is best is up for debate.)

If that's an acceptable workaround for now, I can add something along those lines?

wu-lee: For example: https://stackoverflow.com/questions/14221579/how-do-i-add-comments-to-package-json-for-npm-install

rogup: Yeah sure, looks good

wu-lee:
Comments added...

rogup (approved):

PR #282, opened 2026-04-07
281 powys csv transform

### LINKED ISSUE #281: [SABI] Replace Powys Building Script with CSV Standardisation Script
Currently the powys data is generated by downloading a csv from this spreadsheet: https://docs.google.com/spreadsheets/d/1C8SgDAv0axnivbBoXTPEmqJvoudp4L8F2y5Jaoj7wkw/edit?gid=0#gid=0

And then navigating to `apps/backend` and running `npm run dataset-minimal` or `npm run dataset-minimal-cym` to create the new datasets. This duplicates work when we want to change how the data is generated because it has to happen twice.

What we want is a script that transforms the csv downloaded from google sheets into one ready for the dataset generation script.

Ideally, we have a command that can take the raw google sheet output and generate two csvs, `powys-eng.csv` and `powys-cym.csv` and then you can run the following commands (which use the standard dataset creation script):

`npm run dataset import ./tmp/in/config.en.json ./tmp/in/powys-eng.csv ./tmp/out/powys-eng`

`npm run dataset import ./tmp/in/config.cy.json ./tmp/in/powys-cym.csv ./tmp/out/powys-cym`

The csvs need to be geocoded and otherwise match the config.json property specification.

Once complete, the `datasetMinimal.ts` and `datasetMinimalCym.ts` scripts and corresponding commands in `package.json` can be deleted.

King-Mob:
#### What? Why?

Fixes #281

Changes this so that the code only geocodes and formats a csv for the standard data builder to use.

Design doc: 

#### Checklist

- [ ] Added UTs

- Click '...'


apps/back-end/bin/powysTransform.ts
```diff
@@ -0,0 +1,177 @@
...
+
+const rawEngConfig = await fs.readFile(`./tmp/config.en.json`);
+const rawCymConfig = await fs.readFile(`./tmp/config.cym.json`);
+const engConfig = JSON.parse(rawEngConfig.toString());
+const cymConfig = JSON.parse(rawCymConfig.toString());
+
+// download this data as csv from the google sheet here: 
+// https://docs.google.com/spreadsheets/d/1C8SgDAv0axnivbBoXTPEmqJvoudp4L8F2y5Jaoj7wkw/edit?gid=0#gid=0
+// and name after the date of download
+const rawCSV = (await fs.readFile(`./tmp/2026.04.07_raw_powys.csv`)).toString();
```
laurathackray21: The filename is hardcoded here - could this be passed as an argument instead?

apps/back-end/bin/powysTransform.ts
```diff
@@ -0,0 +1,177 @@
...
+  const food_system_categories = [
+    primary_food_system_category,
+    Object.keys(foodCategories).find(
+      (key) => foodCategories[key] === " - " + item.Subcategory,
+    ),
+  ].join(";");
+
+  const localities = engConfig.vocabs.loc.en.terms;
+      
+      const locality =  Object.keys(localities).find(
```
laurathackray21: Minor thing but some of this isn't formatted very nicely - can you run prettier against it

apps/back-end/bin/powysTransform.ts
```diff
@@ -0,0 +1,177 @@
+import { promises as fs } from "fs";
+import { parse } from "csv-parse/sync";
+import { json2csv } from 'json-2-csv';
+
+const { GEOCODE_TOKEN } = process.env;
+
+const engAbout = (await fs.readFile(`./tmp/about.en.md`)).toString();
+const cymAbout = (await fs.readFile(`./tmp/about.en.md`)).toString();
```
laurathackray21: should this one be about.cym.md?

apps/back-end/bin/powysTransform.ts
```diff
@@ -0,0 +1,177 @@
...
+    const itemOutput = {...item};
+
+    if (item.Geocode) {
+      const [lat, lng] = item.Geocode.split(", ");
+      itemOutput.lat = parseFloat(lat);
+      itemOutput.lng = parseFloat(lng);
+  } else {
+    const cleanAddress = `${itemOutput.Address.replaceAll("\n", ", ")}, ${item.Postcode}, GB`;
+
+    const geocodeResponse = await fetch(
```
laurathackray21: It would be good to add some error handling around this api call

King-Mob:
@laurathackray21 resolved the above comments

docs/data-pipeline.md
```diff
@@ -264,6 +264,12 @@ containing the data in a normalised form, with added or refined
 information (typically geocoded locations). This is typically what we
 use as input to the `dataset import` command.
 
+> [!INFO]
+> We have two ways to generate CSVs for the two projects we support.
```
rogup, line 268: We've almost full standardised our process now apart from this part. The purpose of the https://github.com/DigitalCommons/data-pipelines/tree/main project is for stuff just like this - to take in different datasources and then output a standard.csv that can be consumed by MM's standard script to generate a MM dataset.

So I think that this powysTransform script should go into data-pipelines/apps/powys-data and maybe add a 'download from google sheets' package to data-pipelines/packages that future map deployments can use. Then it will be able to take advantage of all the same deployment work that @wu-lee and @codev are doing

rogup: If you feel like this should happen as a next step once this code has been merged and tested, maybe make an issue to capture the above?

codev: This sounds right to me, I'm working on data/apps/workerscoop and data/apps/dotcoop so it could slot in alongside that

wu-lee: > We've almost full standardised our process now apart from this part. The purpose of the https://github.com/DigitalCommons/data-pipelines/tree/main project is for stuff just like this - to take in different datasources and then output a standard.csv that can be consumed by MM's standard script to generate a MM dataset.

I should point out this isn't quite true, it's more of an aspiration.

The Powys dataset got moved to Typescript - but none of the others datasources have been yet.  And IIRC the Powys dataset got moved in the way it did partly because the week it needed to be done I was on holiday... 

In principle, moving it all to Typescript makes sense in that most people bar me and other other now-departed developers working on open-data were not familiar with Ruby.

But pragmatically, moving all of the existing data normalisation things to TypeScript is a lot of work that will take quite a while. It feels to me that people here are underestimating that.

The code has been around a long time, and it has dealt with a lot of problems and edge cases you get with geocoding, CSVs, and the specific sources we have (ICA, DC, CUK, W.C) all come from very different sources and have to be parsed differently. There's a bunch of stuff to publish the data. Not to mention the RDF generation, vocab mapping, Murmurations publishing code, etc. I expect there's other things too...

Secondly, reinventing that particular wheel with Typescript isn't actually very urgent, because a) it does all currently work, and b) in @codev we have someone who is able and willing to work on Ruby code (I think?)

Thirdly, my experience with processing CSV in TS (in the data-pipelines project) was that it was a bit of a pain. With what seems to be the standard CSV library, it works but with the overhead of a lot of "async call-backery" which make the result a bit harder to work with. Perhaps it's faster? Perhaps we can use dataframe libraries like Pola.rs?  Maybe, but that's going to require quite a lot of R&D, and what we have is adequately fast ATM.

Fourthly, the Powys conversion is in mykomap-monolith - that's not really a good place for it IMO, as it needs to track the Powys project's evolution more than the MM project. It seems a good design principle to decouple MM its upstream data sources, to separate concerns and control complexity. The sensible boundary to decouple on is the interchange of CSV files, which are fairly universally supported and language agnostic.

To conclude: ultimately, I think @codev should decide how he wants to take it forwards.  But I don't think that having everything in Typescript is going to be the big win people seem to assume it will be. And though I say it myself, I am uniquely placed to make that assessment, since no-one currently in DCC except for me has really delved into that codebase.

codev: I'm very happy in Ruby but I don't want to force things to be done twice if this is working and ready to go as I have a lot on my plate already over the next three or four weeks. I'm working on the new config.yaml and trying to get workers coop and dotcoop working without the rest of the CWM map.

I think I'll have a clearer picture once I have a meeting with @wu-lee next week but also who is expected to work on this in future - will it be just me? Or maybe @rogup as well?

rogup: Probably mainly you @codev

I'm not sure how my comment about this Powys script moving to the data-pipelines repo relates to all data processing being converted to TS @wu-lee? I didn't think I was suggesting that. That's probably because I don't understand the full picture of the generation of the dataset CSVs and so don't understand your intention behind data-pipelines and how it relates to TS. Anyway, I guess these are all things mainly for Marc to get to grips with!

wu-lee: Sorry @rogup, apologies for any misunderstanding or apparent rudeness.

In DCC software travails it has been easy to get into a pickle due to miscommunication. I think the JS Powys script the result of one of these miscommunications. This is my attempt to head off more of that, whilst acknowledging that part of the miscommunication comes from threads like this being the primary means of interaction.

What you said above read very much like: the download and normalisation of original data is being standardised to happen in data-pipelines, and moving the Powys script there is just the natural next step.

This also sounded like the way other people saw it in the last delivery meeting, so it feels like it could easily be a common assumption.

rogup: Nws, I didn't think you were being rude.

> the download and normalisation of original data is being standardised to happen in data-pipelines

I hear that this isn't actively happening, since there are lots of Ruby scripts in the open-data project. Things sound like they're in flux.

But is that statement about data-pipelines still a kind of long-term aspiration? Like, data-pipelines being the project that eventually orchestrates the work (which potentially may involve calling Ruby scripts from a different project)?

I feel like some kind of clear 'north star' would be helpful in situations like this, so that the team can understand where we're headed and make more autonomous decisions that align with the long-term vision (e.g. in this case, where to put these new Powys csv generation scripts) without having to have a big discussion each time about exactly what the current state of things is.

wu-lee: I am passing the "baton of north star selection" on to my successors. Also, probably this is entirely off-topic to discuss within this thread (at least in the necessary depth).

PR #294, opened 2026-04-29
Don't clobber the about.md file if it exists

codev:
#### What? Why?

The about.md text is getting replaced in the live version with text saying "created on:" - this is a fallback that should only replace the text if about.md doesn't exist.

### DIFF
--- a/apps/back-end/src/dataset.ts
+++ b/apps/back-end/src/dataset.ts
@@ -232,10 +232,14 @@ export class DatasetWriter {
       await searchableFile.write("\n]}");
 
       // Write a stub about.md file
-      writeFileSync(
-        join(dirPath, "about.md"),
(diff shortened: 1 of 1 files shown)


wu-lee:
I don't think this does any harm... 

But as mentioned on Element: it's not directly the problem, it was me picking the latest (I thought) `about.md` without looking inside. Datasets are generated into a new folder (the CLI checks for non-existance to make sure) so it's pretty much guaranteed there won't be an `about.md` file in there to clobber. 

So I think this branch of the logic will never get triggered.

(The `about.md` has to be added manually afterwards at the moment, although if we go to `config.yaml` files in future as I suggest, this won't be needed any more, because the about text can be included in the YAML.)

So happy to merge or not, and will defer that choice to @codev

PR #329, opened 2026-08-14
321 submap urls

codev:
#### What? Why?

Fixes #326 and #327 in EPIC #321

We want to provide CoMinnesota with their own submap within the CWM. I've added a submap URL parameter that does just that.

config.json now has a section submaps that allows you to lock a filter that is not visible to the user and optionally allows you to set the map bounds, a browser title and some markdown to prefix to the about text:

```
  "submaps": {
    "co-minnesota": {
      "lockedFilter": [
        "data_sources:COMN"
      ],
      "mapBounds": [
        [
          -97.5,
          43.2
        ],
        [
          -89.0,
          49.5
        ]
      ],
      "aboutPrefix": "dataset:assets/co-minnesota-about.md",
      "title": "Co-Minnesota"
    }
  }
```

Initially more results were showing up in the search pane when you searched for nothing so I added some tests to cover that.

#### Checklist

- [x] Added UTs

Run mykomaps with the latest dev cwm-test-data:
- open the root http://localhost:5173
- check that Co-Minnesota appears below CWM
- click on it
- check the map centres on Minnesota
- check only the 4 CoMinnesota points appear (one example one is outside the initial map bounds near 0,0 in the Atlantic)
- check only those appear in a blank search in the results panel
- check the example grocery coop gets a dotcoop verified map pin
- try out search and filters and check they don't show any other points
- check clearing search returns to the CoMinnesota test set and not any other points

#### Deployment notes

Needs to be deployed with data update


docs/config.md
```diff
@@ -213,6 +213,38 @@ The `map` field has 1 subfield
 
 <br />
 
+## Submaps
+
+A submap is a named view of a dataset that only ever shows the items matching a locker filter.
```
laurathackray21: locked filter?

docs/config.md
```diff
@@ -213,6 +213,38 @@ The `map` field has 1 subfield
...
+  "co-minnesota": {
+    "lockedFilter": ["data_sources:COMN"],
+    "mapBounds": [[-97.5, 43.2], [-89.0, 49.5]],
+    "aboutPrefix": "dataset:assets/co-minnesota-about.md",
+    "title": "Co-Minnesota"
+  }
+}
+```
+
+- lockedFilter (required): prop:value filters ANDed into every search. This is removed from the filter UI and applied if the user clears the search or resets the filters and it never appears in the shareable q URL param
```
laurathackray21: If it never appears in the shareable URL query param , does this mean that you can't share a link to the map with this locked filter?

laurathackray21 (approved):
I've reviewed this the best I can - some of the existing searching logic is hard to read/understand 😅 
I've tested it all out though and it works really well! I couldn't get it to show any other data points than the 4 example ones.

For my broader understanding - what is the reason for the submap rather than them having a standalone map?

docs/config.md
codev: No you can, I mean the query never appears in the q param, it appears in the submap param

codev: Added a small note on that

codev:
We went with a submap instead of a new map so you can have maps for each of the CWM data sources/APEX maps and they will never get out of sync - if we generated a whole new map for CoMinnesota we'd have to:

1. make sure every time another data source that updates something in CoMinnesota was generated (for example if dotCoop verify a domain) both the CWM and CoMinnesota maps would have to be regenerated - this way each point only has one .json file
2. generate a new config for just CoMinnesota and keep it up to date

PR #331, opened 2026-08-26
Cloudron packaging and Dockerfile cleanup

codev:
Part 1 of technology-and-infrastructure#173 - FE reads API keys from a runtime config.js; adds CloudronManifest.json and cloudron (with the startup script, Dockerfile and Caddyfile) building a single image on top of cloudron/base instead of two images. Stacked, will add PRs for CI and building fo rthe GHCR, CLoudron catalogue settings and the dataset upload


apps/front-end/public/config.js
```diff
@@ -0,0 +1,27 @@
+// MykoMaps config
+//
+// Fill this in before using the app to make features work
+// MAPTILER_API_KEY is required to display the map
+// The others are optional
+//
+// Get a MapTiler key: https://docs.maptiler.com/cloud/api/authentication-key/
+//
+window.MYKOMAP_CONFIG = {
```
laurathackray21, line 9: This seems like a non standard way of doing config? What's the reason for this over using env vars?

apps/front-end/src/config.ts
```diff
@@ -0,0 +1,64 @@
+// FE runtime configuration
+//
+// Values come from config.js - see apps/front-end/public/config.js
```
laurathackray21: I don't understand why we're doing this - why aren't we just using env vars?

apps/front-end/public/config.js
```diff
@@ -0,0 +1,27 @@
...
+   //// Get it: Settings -> Projects -> MykomapFrontEnd (create this)
+   ////   -> Security Endpoint URL
+   //GLITCHTIP_KEY: "abc123",
+   //// Umami analytics - you can self host and change the host
+   //UMAMI_URL: "https://umami.is/script.js",
+   //UMAMI_ID: "00000000-0000-0000-0000-000000000000",
+   //// Optional Umami session recorder (heatmaps and replays)
+   //UMAMI_RECORDER_URL: "https://umami.is/recorder.js",
+   //// Mixpanel analytics token and session recording percentage (0 to 100)
+   //MIXPANEL_TOKEN: "abc123",
```
laurathackray21, line 25: Are all these keys safe to expose in the frontend?

apps/front-end/public/config.js
laurathackray21: Ah I see now from reading the rest of the PR it's so people can use their own keys for analytics etc.

laurathackray21 (commented):
Reviewed the best I can - My main query is over env vars being exposed. I've realised from reviewing this that we shouldn't be prefixing any API keys with `VITE_`
https://vite.dev/guide/env-and-mode as this exposes them client side. Might need to review LandExplorer for this too.

laurathackray21:
Also, I think it would be good at some point to do a quick overview of cloudron with Marcel and I :)

apps/front-end/public/config.js
codev: The keys were all public anyway - when the map loads tiles from MapTiler the browser makes the request. Same for GlitchTip, Umami and Mixpanel - browser does the requests.

VITE_ means put them in the bundle so anyone could extract them from the JS or network requests.

However it has made me think we need to check if MapTiler is domain restricted - will do that now.

apps/front-end/public/config.js
codev: Yes - the front end is a static React app served by Caddy - previously we baked the vars into the build but we can't do this if we're packaging it as an app.

codev:
The variables were always exposed on the FE, this just brought them together as the existing ones were baked into the code at build time which doesn't work if you need to change them between stages/for different users.

I checked the MapTiler key and it wasn't restricted by domain - now added *.digitalcommons.coop *.landexplorer.coop and *.maps.coop - works with iframe embedding (checked CWM and Powys) without adding their domains.

Created an issue for LandExplorer: https://github.com/DigitalCommons/land-explorer/issues/170

